PR #3, opened 2017-02-09
Create README.md

netroby:


netroby:
@xrme have a look?

xrme:
Thanks for the enthusiasm.  A README.md file is an excellent suggestion.

But rather than marketing text, I think it should include instructions on how to fetch everything needed (sources, bootstrapping binaries, interface databases) to build a new copy of CCL.  And, to be honest, I haven't yet figured out how that will end up working.  I hope to figure that out this weekend.  Perhaps GitHub releases will be suitable.

netroby:
I wang to help. so i first create the README.md, then  i would like continue to improve the document.

xrme:
I added a README.md file focused on describing how to build the sources.  I think it's better to leave the historical and marketing text to http://ccl.clozure.com/.

Thanks for your enthusiasm and willingness to help.

PR #6, opened 2017-02-12
add object files to the git ignore

eval-apply:
Small change to exclude object files from the git tree


xrme:
Thanks.  That `.gitignore` file is kind of mess; it was generated from svn:ignore properties.

PR #18, opened 2017-02-28
gitignore: ignore some more ephemera

tfeb:
This tries to make it be the case that `git status` on a repo where the release tarball has been unpacked and a build has been done but no other changes have been made will be completely quiet.

In particular ignore the darwin headers directories which come from
the release tarballs, and additionally ignore some object files in the AltConsole, and some
extra places where fasls seem to end up when rebuilding.

Note: this has only been checked on a Mac: there may be similar changes needed for other platforms.  Sorry.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 bin/*fsl
 cocoa-ide/*fsl
 cocoa-ide/*~.*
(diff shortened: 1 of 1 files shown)


xrme:
Thank you.

PR #29, opened 2017-03-18
In %bignum-random, normalize generated bignum

gzacharias:
%bignum-random wasn't normalizing its bignums before passing them to MOD (aka REM).

Luckily the current implementation of REM managed to ignore the bogosity and happened fix up the result, at least in the 64 bit version.  Perhaps 32 bit version wasn't as lucky -- see ticket 1403 in trac.  Note however that this change is in l0-bignum64 only.  I can't find the 32 bit version of %bignum-random, so I couldn't fix it.


xrme:
The 32-bit %bignum-random is at https://github.com/Clozure/ccl/blob/master/level-0/l0-numbers.lisp#L1891

Issue #38, opened 2017-05-13
determine_executable_name doesn't do the job for FreeBSD


xrme:
The function `determine_executable_name()` in the lisp kernel is supposed to return the full path to the running lisp kernel executable.

This function uses various system-specific ways to do this, depending on what the platform the lisp kernel is being compiled for.  Unfortunately, the bit of code for FreeBSD doesn't do the job.  It essentially just calls `realpath(3)` on the value of `argv[0]`.

To see why this is inadequate:
```
$ ./fx86cl64
? (save-application "magic" :prepend-kernel t)
$ cp magic ~/bin    # or some other place in your path
$ cd /
$ magic
Couldn't load lisp heap image from magic.image: No such file or directory
```

This problem was discovered in https://github.com/dimitri/pgloader/issues/551

xrme:
Also merged to the 1.11 branch in fa63981

PR #39, opened 2017-05-13
In determine_executable_name, use sysctl(3) on FreeBSD

xrme:
Just calling realpath(3) on the value of argv[0] is inadequate.

See #38.


xrme:
Also merged to the 1.11 branch in fa639816

Issue #44, opened 2017-06-07
CCL:ARGLIST interns symbols in the current package


eugeneia:
From https://trac.clozure.com/ccl/ticket/1260

> SLSIA:

```
Welcome to Clozure Common Lisp Version 1.10-r16304M  (DarwinX8664)!
? (find-symbol "ARGS")
NIL
NIL
? (arglist 'setf)
(&REST ARGS)
:DECLARATION
? (find-symbol "ARGS")
ARGS
:INTERNAL
```

> The problem occurs only when calling CCL::ARGLIST on macros, not functions.

> I know it's not kosher to complain about the behavior of unexported functions, but swank uses this and it causes real problems. For example, trying to (use-package :cl-who) in a swank session fails because the symbol STR gets interned in cl-user during the swank startup process.

First of all I want to note that `ccl:arglist` is indeed external to the `ccl` package, and seems to be the only way to get lambda lists of functions/macros/..., which is a common operation in introspection tools (SLIME, documentation generators, ...). I would argue this to be a popular feature, and hence a problematic bug.

I’ve investigated a bit, and the interning happens because for macros the lambda list is saved as a string (returned by `%arglist-internal`) which is then parsed by `ccl::%arglist` using `read` in the current package.

In a way this is not a bug since its not defined which package should intern the symbols in question.

xrme:
fixed by #47

PR #45, opened 2017-06-07
ccl:arglist - set explicit *PACKAGE* during READ

eugeneia:
ccl:arglist would intern symbols of macro lambda lists into the current
package. This patch changes the current package to the SYMBOL-PACKAGE
of the first argument.

I.e.: (ccl:arglist 'cl:setf) → (&REST COMMON-LISP::ARGS)

See Also: #44

### DIFF
--- a/lib/arglist.lisp
+++ b/lib/arglist.lisp
@@ -82,7 +82,8 @@
           (declare (dynamic-extent eof))
           (loop
             (multiple-value-setq (val errorp)
-              (ignore-errors (values (read stream nil eof))))
+              (ignore-errors (values (let ((*package* (symbol-package sym)))
+                                       (read stream nil eof)))))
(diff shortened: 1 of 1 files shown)


xrme:
I have to admit that I don't know what to do here.

Is interning symbols into, e.g., the `CL` package in particular is any better than what we are currently doing?

gzacharias:
Why intern them at all?

eugeneia:
I think not interning at all would be the neatest solution. However I can’t think of a way to do that with `read`.

I also agree with you @xrme that this is not more correct than the status quo, merely what other implementations seem to do. Maybe it would be more appropriate to submit patches to SLIME etc.

eugeneia:
Superseded by #47

Issue #20, opened 2017-03-03
get-internal-real-time should return monotonic time


eugeneia:
I think `get-internal-real-time` should return a monotonic/non-decreasing value. Even though the function is not mentioned in the manual (also an issue), its docstring says

> Return the real time in the internal time format. (See INTERNAL-TIME-UNITS-PER-SECOND.) **This is useful for finding elapsed time.**

which a non-monotonic clock is not. Proposed solution: have `get-internal-real-time` derive its return value from `clock_gettime(3)` instead of `gettimeofday(2)`.

xrme:
There's `ccl:current-time-in-nanoseconds`, which was implemented in response to http://trac.clozure.com/ccl/ticket/924.

I am not against having `get-internal-real-time` use the same mechanism.  However, I understand that some of these monotonic clocks may stop increasing during system sleep.

Linux, FreeBSD, and Solaris-ish systems have `clock_gettime`.

The Macintosh has `mach_absolute_time`: https://developer.apple.com/library/content/qa/qa1398/_index.html

For Windows, there's `QueryPerformanceCounter`:
https://msdn.microsoft.com/en-us/library/windows/desktop/dn553408(v=vs.85).aspx

https://www.python.org/dev/peps/pep-0418/ collects a lot information about various systems.

eugeneia:
According to [PEP 418  / Operating system time functions / Monotonic Clocks](https://www.python.org/dev/peps/pep-0418/#monotonic-clocks) the options you mentioned all do include sleep. Its looks though, as if `current-time-in-nanoseconds` does not deliver on windows where it uses `GetSystemTimeAsFileTime`, which according to [MSDN documentation](https://msdn.microsoft.com/en-us/library/windows/desktop/ms724397%28v=vs.85%29.aspx) (also see “System Time”) is non-monotonic:

>  Retrieves the current system date and time.

> System time is the current date and time of day. […] Note that the system can periodically refresh the time by synchronizing with a time source. Because the system time can be adjusted either forward or backward, do not compare system time readings to determine elapsed time.

Disclaimer: I am absolutely clueless of windows APIs.

`get-internal-real-time` could be something like this

```
(defun get-internal-real-time ()
  (nth-value 0 (floor (current-time-in-nanoseconds)
                      (load-time-value
                       (/ 1000000000 internal-time-units-per-second)))))
```

but I am afraid I would be no help at all with the windows implementation of `current-time-in-nanoseconds`.

[...]

PR #46, opened 2017-06-08
[WIP] Make get-internal-real-time monotonic

eugeneia:
See Also: #20

This work-in-progress PR moves `ccl:current-time-in-nanoseconds` from lib/time to l1-lisp-threads, and uses it to implement `get-internal-real-time` and `ccl::get-tick-count`.

The previous implementation of `get-internal-real-time` remembered the initial seconds value and subtracted it from values before returning them. I.e. the values returned by `get-internal-real-time` would be close to zero after starting CCL. The new implementation does no such thing, the value returned is relative  to system boot time.

Additionally it changes `ccl:current-time-in-nanoseconds` to

- return a monotonic value on Windows (>=Vista, see https://msdn.microsoft.com/en-us/library/windows/desktop/ms724411%28v=vs.85%29.aspx)
- return a nanosecond value on Darwin (presumably, there was a bug here, see https://developer.apple.com/library/content/qa/qa1398/_index.html)

Could anyone on Windows/Darwin test these changes? I remember Microsoft to provide virtual machine images for testing, but I doubt they include a C compiler? Anyway, this should print as shown below:

```
(let* ((a (get-internal-real-time)) (at (ccl::get-tick-count))
         (_ (sleep 3))
        (b (get-internal-real-time)) (bt (ccl::get-tick-count)))
   (print (float (/ (- b a) internal-time-units-per-second)))
   (print (float (- bt at)))
   (print ccl::*ticks-per-second*)
   (values))
;Compiler warnings :
;   In an anonymous lambda form at position 0: Unused lexical variable _
3.000157 
3000.0 
1000
```

### DIFF
--- a/level-1/l1-lisp-threads.lisp
+++ b/level-1/l1-lisp-threads.lisp
@@ -96,12 +96,6 @@
     (* (logior (pref time #>FILETIME.dwLowDateTime)
                (ash (pref time #>FILETIME.dwHighDateTime) 32))
        100)))
-(defloadvar *lisp-start-timeval*
-    (progn
-      (let* ((r (make-record :timeval)))
-        (gettimeofday r)
-        r)))
-
 
 (defloadvar *internal-real-time-session-seconds* nil)
 
--- a/level-1/l1-lisp-threads.lisp
+++ b/level-1/l1-lisp-threads.lisp
@@ -76,6 +76,26 @@
                     :address (or ptz (%null-ptr))
                     :int))
 
+#-(or darwin-target windows-target)
+(defloadvar preferred-posix-clock-id
+  (rlet ((ts :timespec))
+    (if (eql 0 (#_clock_gettime #$CLOCK_MONOTONIC ts))
+      #$CLOCK_MONOTONIC
+      #$CLOCK_REALTIME)))
+
+(defun current-time-in-nanoseconds ()
+  #-(or darwin-target windows-target)
+  (rlet ((ts :timespec))
+    (#_clock_gettime preferred-posix-clock-id ts)
+    (+ (* (pref ts :timespec.tv_sec) 1000000000)
+       (pref ts :timespec.tv_nsec)))
+  #+darwin-target (#_mach_absolute_time)
+  #+windows-target
+  (rlet ((time #>FILETIME))
+    (#_GetSystemTimeAsFileTime time)
+    (* (logior (pref time #>FILETIME.dwLowDateTime)
+               (ash (pref time #>FILETIME.dwHighDateTime) 32))
+       100)))
 (defloadvar *lisp-start-timeval*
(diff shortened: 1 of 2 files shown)


level-1/l1-lisp-threads.lisp
```diff
@@ -33,8 +33,17 @@
     #-windows-target
     (max 1000 (#_sysconf #$_SC_CLK_TCK)))
 
+(defloadvar *ns-per-second*
+    1000000000)
```
xrme: This looks like a job for defconstant.

level-1/l1-lisp-threads.lisp
```diff
@@ -33,8 +33,17 @@
...
+
+(defloadvar *ns-per-millisecond*
+    (floor *ns-per-second* 1000))
```
xrme: Also defconstant.

xrme:
For the record, I believe the reason that we saved the initial seconds value from gettimeofday was so that operations on internal time unit values would likely be fixnum operations.

It's definitely more correct to use `mach_timebase_info` instead of assuming that Mach absolute time units are nanoseconds.  They are on the Mac, but I think they're different on iOS devices (not that there's much chance that we'll run on iOS devices, but you never know).  Good catch there.

If we use `GetTickCount64`, that means we're dropping support for running on Windows XP.  I'm not necessarily against this, but we need to make that clear.

We might also have to generate new Windows headers so that the `#_` reader macro will know about `GetTickCount64`.  Or we could just use `ccl:external-call` and put off dealing with new headers.  I'll check all this on a Windows box as soon as I can.

eugeneia:
Note that only on Windows this `ccl:current-time-in-nanoseconds` includes the time passed while the system is suspended. For instance, the laptop lid is closed, etc.

eugeneia:
I also thought about fixnums with regard to `internal-time-units-per-second`.

```
(float (/ most-positive-fixnum internal-time-units-per-second 60 60 24))
```

- On 32bit CCL the value returned by `get-internal-real-time` is a fixnum for at most 6.2137837 days
- On 64bit CCL the value returned by `get-internal-real-time` is a fixnum _forever_
- If `internal-time-units-per-second` were nanoseconds the value returned by `get-internal-real-time` is a fixnum for at most ~36 years on 64bit CCL and always boxed on 32bit CCL

I am temped to ignore the six days of fixnums on 32bit CCL because it’s an easily imaginable run time of a program. Since we are getting time in nanoseconds here anyways, the 36 year mark applies indirectly through `ccl:current-time-in-nanoseconds`. I figured we _could_ change `internal-time-units-per-second` to nanosecond precision. It’s a somewhat messy change (I ended up needing to rebuild twice to get a correct build because the compiler did constant things), but would simplify things.

PR #47, opened 2017-06-09
defmacro - record fully qualified argstring

eugeneia:
ccl:arglist would have the unexpected side effect of semi-randomly
interning symbols found in macro lambda lists. This change ensures that
symbols are only interned in packages where they have actually been used.

See http://lisptips.com/post/20871695085/printing-package-qualified-symbols

See Also: #44

### DIFF
--- a/lib/macros.lisp
+++ b/lib/macros.lisp
@@ -670,7 +670,8 @@
                           (if (eq arg '&aux)
                             (return)
                             (push arg temp)))
-                        (format nil "~:a" (nreverse temp)))))
+                        (let ((*package* (find-package "KEYWORD")))
+                          (format nil "~:S" (nreverse temp))))))
       (if (and body-pos (memq '&optional normalized)) (decf body-pos))
       `(progn
          (eval-when (:compile-toplevel)


eugeneia:
On a different note, why does `defmacro` record the lambda list as a string in the first place? The only thing I could think of was maybe to take load of the GC? But since `defun` and `defgeneric` seem to record the lists directly…

gzacharias:
I assume it has something to do with packages, perhaps packages that exist that compile time not existing at runtime.  Note that defmacro specifically uses ~a rather than ~s to print the arglist.  Perhaps the safest solution would be to walk the arglist and replace all non-CL-package symbols with uninterned copies.

eugeneia:
That explanation makes way more sense of course!

eugeneia:
Why treat CL-package symbols special though? I am thinking it would clash with using `ccl:arglist` in packages that don’t use `:cl`?

gzacharias:
It wouldn't clash, but now that you mention it, it would look weird.  It should special-case lambda-list-keywords only, rather than all CL symbols.

eugeneia:
I still think this is a correct solution. Symbols that can not be read at run-time (i.e. because a package is not present that existed at compile-time), will trigger the `:unparsable` bailout in `ccl::%arglist` which I think is a sensible way to default in this rare situation.

The downside with non-interned symbols is that they don’t actually reflect what you want to get when you call `ccl:arglist` (which is “give me the list that was the lambda list argument to `defmacro`”). On the other hand, if I understand correctly, it isn’t actually always possible to get _that excact thing_, and in cases where you can’t, I believe its better to bail out than present lossy information. Make sense?

What do you think @xrme ?

xrme:
This seems acceptable.  Thank you, @eugeneia.

eugeneia:
I celebrate my first source contribution to CCL! :fireworks: :champagne: :beers:

PR #48, opened 2017-06-09
defmacro - record lambda list as list

eugeneia:
This removes the special case code in place to record lambda lists of macros in printed from, for reasons unknown to me. This is a possible alternative to #47.

See Also: #44


eugeneia:
https://github.com/Clozure/ccl/pull/47#issuecomment-307469524

Issue #49, opened 2017-06-10
typo: #ifdef WINDOXS


xrme:
Typo at https://github.com/Clozure/ccl/blob/master/lisp-kernel/pmcl-kernel.c#L1932

"WINDOXS" should be "WINDOWS"

The fix needs to be compiled and tested on Windows.

PR #56, opened 2017-07-07
fix typo

ailisp:
In https://github.com/Clozure/ccl/issues/49
Tested build in Windows 7 64 bit, I build both 32 bit and 64 bit ccl there.


xrme:
Thank you.

Issue #55, opened 2017-07-06
Windows build wx86cl64 need libwinpthread-1.dll


ailisp:
When building kernel, wx86cl64 complains missing libwinpthread-1.dll. 32bit wx86cl is fine. There is no explicit reference to this in windows Makefile. Do a search found that this is required by libgcc, which is  required by mingw-w64 x86_64 build. We can make it statically linked. I'll send a pull request for this soon.

PR #57, opened 2017-07-07
fix need libpthread-1.dll in windows 64 build

ailisp:
See https://github.com/Clozure/ccl/issues/55
Checked with visual studio's dumpbin.exe. 32 bit ccl doesn't require libpthread-1.dll. But 64 bit ccl does. It's added by mingw64-x86_64-gcc. After static link it, now it's able to run wx86cl64 with only ccl image
Reference:
https://stackoverflow.com/questions/13768515/how-to-do-static-linking-of-libwinpthread-1-dll-in-mingw

### DIFF
--- a/lisp-kernel/win64/Makefile
+++ b/lisp-kernel/win64/Makefile
@@ -97,7 +97,7 @@ KSPOBJ = $(SPOBJ)
 all:	../../wx86cl64.exe
 
 
-OSLIBS = -lpsapi -lws2_32
+OSLIBS = -lpsapi -lws2_32 -static -lpthread
 
 
 ../../wx86cl64.exe:	$(KSPOBJ) $(KERNELOBJ) $(DEBUGOBJ) Makefile pei-x86-64.x


xrme:
Does it work simply to add `-lpthread` to `OSLIBS` in both the win32 and win64 makefiles?

On my system (gcc/i686-w64-mingw32/4.9.2), this works fine.  The `-lpthread` doe not seem to be required, but its presence appears to cause no trouble.

I assume that you are using a more recent version, where the`-lpthread` is required.  If so, let's just add `-lpthread` to `OSLIBS`.  I'd prefer to avoid having to fiddle with static linking if that is possible.

ailisp:
My gcc is i686-w64-mingw32-gcc-5.4.0.exe. Just installed in cygwin's setup about a week ago. If you open cygwin installer, and forward to the package selection step, it will ask you to upgrade to latest version. 
For `-static-link`, unfortuately, if I remove it, after build:
```
C:/cygwin64/home/Administrator/ccl/wx86cl64.exe: error while loading shared libraries: libwinpthread-1.dll: cannot open shared object file: No such file or directory
```
And the dumpbin shows:
```
C:\cygwin64\home\Administrator\ccl>dumpbin /dependents wx86cl64.exe
Microsoft (R) COFF/PE Dumper Version 9.00.30729.01
Copyright (C) Microsoft Corporation.  All rights reserved.


Dump of file wx86cl64.exe

File Type: EXECUTABLE IMAGE

  Image has the following dependencies:

    ADVAPI32.dll
    KERNEL32.dll
    msvcrt.dll
    PSAPI.DLL
    libwinpthread-1.dll
    SHELL32.dll
    WS2_32.dll
```
So it's still dynamic linked. If `-static` added, the dumpbin output will be:
```
C:\cygwin64\home\Administrator\ccl>dumpbin /dependents wx86cl64.exe
Microsoft (R) COFF/PE Dumper Version 9.00.30729.01
Copyright (C) Microsoft Corporation.  All rights reserved.


Dump of file wx86cl64.exe

File Type: EXECUTABLE IMAGE

  Image has the following dependencies:

    ADVAPI32.dll
    KERNEL32.dll
    msvcrt.dll
    PSAPI.DLL
    SHELL32.dll
    WS2_32.dll
```
And the wx86cl64.exe size increases from 1190KB to 1321KB. I'm afraid we need to keep this `-static`. And for 32bit, wx86cl.exe can run without libpthread-1.dll (and it's not in its dependencies list), so only added in 64 bit.

xrme:
OK.  Thank you for the detailed explanation.  I don't like this situation, but it appears that there's no real alternative.  We certainly don't want to require users to have `libwinpthread-1.dll` installed on their systems.

https://sourceforge.net/p/mingw-w64/mailman/message/31213279/

This makes #59 seem like an even better idea.

PR #58, opened 2017-07-08
building in both cygwin and msys2

ailisp:
Msys2 is now very popular and I saw there are lispers comments in issues that they also use msys2 instead of cygwin. Compare to cygwin, it has pacman as package management tool, but less POSIX support. And msys2 is official recommended way to install gnu emacs on windows, so I believe there're many lisp programmers using msys2. For building ccl in msys2, it doesn't have many difference than cygwin so I modify Makefile and provide a instruction in comment. Tested on both cygwin and msys2 after modification. Because https://github.com/Clozure/ccl/pull/57 is still under discussion, currently only modify win32 Makefile.

### DIFF
--- a/lisp-kernel/win32/Makefile
+++ b/lisp-kernel/win32/Makefile
@@ -44,21 +44,18 @@ SVN_REVISION := "$(shell svnversion || echo unknown)"
 
 VPATH = ../
 RM = rm
-CC = i686-w64-mingw32-gcc
 
 ifeq ($(MSYSTEM),)
+CC = i686-w64-mingw32-gcc
 AS = i686-w64-mingw32-as
+LD = i686-w64-mingw32-ld
 else
+CC = /mingw32/bin/i686-w64-mingw32-gcc
 AS = /mingw32/bin/as
+LD = /mingw32/bin/ld
 endif
 
 M4 = m4
-
-ifeq ($(MSYSTEM),)
-LD = i686-w64-mingw32-ld
-else
-LD = /mingw32/bin/ld
-endif
 
 ASFLAGS = -g --32
 M4FLAGS = -DWIN_32 -DWINDOWS -DX86 -DX8632
--- a/lisp-kernel/win32/Makefile
+++ b/lisp-kernel/win32/Makefile
@@ -14,8 +14,9 @@
 # limitations under the License.
 
 # This makefile is written to be used with the mingw toolchain
-# running in the Cygwin environment.
-#
+# running in the Cygwin environment or Msys2 environment.
+
+# If running in the Cygwin environment:
 # Install the following Cygwin packages:
 # * make
 # * m4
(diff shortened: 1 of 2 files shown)


xrme:
This seems fine.  Let's take care of #57 first, and then you can add an updated win64 makefile to this PR, and I'll merge it.

ailisp:
Thank you for review. It's now successfully build in both cygwin and msys2. However, the `-static -lpthread` is still annoying. In both cygwin and msys2 64bit build requires libwinpthread-1.dll without `-static`. I did some searching find libwinpthread is for support c++ std::thread, which we obviously not needed:
https://bugs.debian.org/cgi-bin/bugreport.cgi?att=0;bug=748353;msg=10

This dependence was at least exist in 4.9.x times. Now new cygwin installation will install gcc 5.4.0 and msys2 will install 6.3.0. Because we only use C, we can use a mingw64 win32 thread build:
https://stackoverflow.com/questions/17242516/mingw-w64-threads-posix-vs-win32
Note this version of mingw is not in msys2 and cygwin package list and can only manually download from sf:
https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win64/Personal%20Builds/mingw-builds/6.3.0/

I tested using this one 
https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win64/Personal%20Builds/mingw-builds/6.3.0/threads-win32/sjlj/x86_64-6.3.0-release-win32-sjlj-rt_v5-rev2.7z/download
to replace `msys2/mingw64` dir, delete `-static -lpthread`, running make in `msys2-dir/mingw64.exe` (not `msys2-dir/msys2.exe`) and got a `wx86cl64.exe` without depends on `libwinpthread-1.dll`.

I think we have three choices:
1. keep using old version mingw;
2. add `-static -lpthread` for 64bit make file;
3. require windows developer using this win32-threads build mingw64. (This is 

If we use 3, we may need to reorganize mingw files to build in cygwin, because it hasdifferent file organization for mingw64. The msys2 organize mingw64 files exactly the same as uncompress mingw64, but cygwin put them all in `/bin` and rename. What do you think?

xrme:
Building the lisp kernel on Windows has, unfortunately, always been somewhat difficult due to toolchain issues like this.

I think the goal for now should be for users to be able to install Cygwin (or msys2) and the packages that are listed in the Makefile comments, and be able to build a working, redistributable lisp kernel binary.  Since the win32 thread build of mingw64 isn't packaged, I don't think that option 3 is suitable.

In some ways, I'd like to stick to option 1, but that suffers from the same problem:  users now have to go to special effort in order to obtain and install a non-default version of the mingw64 toolchain.

My opinion, then, is that option 2 is the best choice at the moment, even though I don't like it.

ailisp:
Yes. It's quite inconvenient to program on windows, compare to mac and
linux. Since you prefer option 2, now this PR is done in this way for 32
bit and 64 bit build in cygwin and msys2. Please have a look, thank you.

PR #62, opened 2017-07-28
Make FD-STREAM-EOFP reliably return T when at EOF.

jathd:
Right now, `fd-stream-eofp` just checks the `iobuffer-eof` flag. This is wrong because the only function that sets this flag is `fd-stream-advance`, which isn't called unless we actually try to read from the stream.

For instance, open an empty file and immediately call `stream-eofp`: you get NIL.


xrme:
Thank you.

PR #83, opened 2017-11-16
Bugfix to ccl::basename.  Specifying a directory now yields a relativ…

jkordani:
…e pathspec instead of an absolute, i.e. #P"/one/two/three/" ->> #P"three/" instead of #P"/three/"

This breaks ccl::path, if one of the arguments to path is an absolute directory path, then the resulting path value is incorrect.

This breaks ccl::copy-nibfile, if srcnib is a directory into which we should recurse, its target output directory is incorrectly constructed (to a directory in root, no less!)


xrme:
Thanks for the first-time PR!

jkordani:
Thank you!

PR #87, opened 2017-11-22
Some error-handling fixes

svspire:


lib/backtrace.lisp
```diff
@@ -647,6 +647,7 @@ object."
                        (dolist (r rest) (push r result)))
                      (return))
                     (&rest
+                     (unless (listp value) (setf value (list value)))
```
xrme: How can a `&rest` argument be anything other than a list?  That seems pretty strange.

svspire: Look at #'arguments-and-locals. If #'map-entry-value ever returns a non-list, then #'arguments-and-locals can return a list that contains a dotted pair--the cdr of which will be a non-list. Am I reading this right? That could explain how it happens.

PR #94, opened 2017-12-21
Expose Lisp thread names to Linux

ruricolist:
On Linux only, this patch uses `pthread_setname_np` to make the names of Lisp threads visible to standard Linux tools (e.g. GDB, htop).

I hacked this in for myself, but I thought it might be useful for others.

### DIFF
--- a/level-1/l1-lisp-threads.lisp
+++ b/level-1/l1-lisp-threads.lisp
@@ -463,7 +463,35 @@
         (unless (zerop natural) natural)))))
 
 
-                         
+
+#+linux-target
+(defun set-os-thread-name (lisp-thread name)
+  "Set the OS-visible name of the thread with pthread_setname_np."
+  (let ((name
+          (if (<= (length name) 15)
+              name
+              (subseq name 0 15))))
+    (let ((natural (lisp-thread-os-thread lisp-thread)))
+      (when natural
+        (with-utf-8-cstr (name name)
+          (external-call "pthread_setname_np"
+                         :pthread_t natural
+                         :address name
+                         :int))))))
+
+#+linux-target
+(defun get-os-thread-name (lisp-thread)
+  "Read the OS-visible name of the thread with pthread_getname_np."
+  (let ((natural (lisp-thread-os-thread lisp-thread)))
+    (when natural
+      (%stack-block ((buffer 16))
+        (external-call "pthread_getname_np"
+                       :pthread_t natural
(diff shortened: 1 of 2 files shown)


level-1/l1-lisp-threads.lisp
```diff
@@ -463,7 +463,35 @@
...
+#+linux-target
+(defun set-os-thread-name (lisp-thread name)
+  "Set the OS-visible name of the thread with pthread_setname_np."
+  (let ((name
+          (if (<= (length name) 15)
+              name
+              (subseq name 0 15))))
+    (let ((natural (lisp-thread-os-thread lisp-thread)))
+      (when natural
+        (with-utf-8-cstr (name name)
```
xrme: Note that (length name) is not, in general, guaranteed to be the same as (string-size-in-octets name :external-format :utf-8).

According to pthread_setname_np(3), it will return ERANGE if the string is too long. I suppose it's not the end of the world if the call just fails, but I'd like to see us ensure that the null-terminated encoded utf-8 data fits in 16 octets.

ruricolist: Is string-size-in-octets usable at level 1? I tried introducing it but reloading the boot image fails with "Can't find symbol".

xrme: You are right about string-size-in-octets.  However, you should be able to use ccl::utf-8-octets-in-string, which is loaded in level-0, and should be available.

If that's too much of a pain and you don't feel like dealing with the variable-length utf-8, you could use with-cstr and %get-cstring.  Those would use iso-8859-1.

I think this a nice little addition; I hope that you don't mind having to deal with my comments.

eugeneia: Since I am interested in being able to spawn as many threads as possible, as quickly as possible, I would want to minimize the consing involved in thread creation. I.e. ideally for it to be constant time.

So I guess Linux guarantees you ASCII support, does it support a specific encoding like utf-8 or iso-8859-1 though? I can’t immediately tell from PTHREAD_SETNAME_NP(3). Maybe make a C string out of the first 15 ASCII characters, ideally without consing vita `subseq`?

ruricolist: I've updated the pull request to use with-cstr and %get-cstring, so there should be no consing.

xrme:
Thank you, @ruricolist.

Issue #59, opened 2017-07-08
use Microsoft tools to build lisp kernel [$700]


xrme:
I think it would be worthwhile to make the lisp kernel buildable with Microsoft's toolchain (either or both of the command-line tools or with Visual Studio).

My understanding is that this would let us to Microsoft's debugging tools (such as windbg), which may be better (in some sense) than using gdb.

One challenge is going to be dealing with the various assembly language files.  There's an assembler called yasm (http://yasm.tortall.net) which should be able to support the AT&T style syntax that we use.

xrme:
To elaborate on this a little bit more:

On Windows, we build the lisp kernel with Cygwin or msys2.  Please see the comments at the top of the Makefile in the appropriate lisp kernel build directory (e.g., https://github.com/Clozure/ccl/blob/master/lisp-kernel/win64/Makefile).

As mentioned already, the main task to building with some other toolchain is going to be building the assembly-language files.  The `.s` files are processed with m4, since system assemblers have historically been pretty weak in terms of supporting macro facilities.

If we can come up with some way to do this in the context of building with Microsoft's toolchain, this might also make it easier to build the lisp kernel with Xcode, as the linked issue mentions.

ailisp:
I got some time this weekend, will take a look. And if this comment also post on bountysource will show bountysource and github linking is suceessful.

xrme:
See @ailisp's PR #139

denim2x:
@xrme [Here](https://github.com/denim2x/ccl/wiki/Lisp-kernel:-compatibility-with-WinDbg) are some possible approaches for solving this; please let me know what you think

xrme:
@denim2x I think using clang would be fine.  There is no requirement to use Microsoft's compiler.  Just as a reminder, the idea behind all of this is to be able to use windbg and possibly other Microsoft debugging tools instead of using gdb running under Cygwin.

The trouble has always been the assembly-language files.  The first challenge is running them through m4 as part of the build process.  I know there are Windows versions of GNU m4 out there.

Is there an assembler that goes along with the clang/LLVM toolchain?  If so, and if it supports the AT&T syntax, then that's probably the way to go.

denim2x:
[...]

PR #139, opened 2018-07-15
[WIP] Build by Visual Studio Community 2017

ailisp:
Current status: can generate lisp-kernel.exe (wx86cl64.exe) on a Windows 7 x64 computer, but don't working. With error message:
minimal initial mmap: 0x8 (and a following Chinese sentence because of my windows version, these characters mean "Insufficient storage space, unable to process this command")
 Not sure how much work will need to solve this.

Steps to build:
1. Install Visual Studio Community 2017,
2. Install  VSYASM from here: https://github.com/ShiftMediaProject/VSYASM
3. Install Msys2, then install m4 and mingw-w64-x86_64-gcc package (for using GNU as, yasm can't build some of the .s file, like x86-asmutils64.s, has several error messages like `x86-asmutils64.s.asm:3542: error: undefined symbol `win64_context' in preprocess`. Since anyway we can't use microsoft asm (MASM), maybe here use GNU AS instead of YASM is fine? (MS link.exe can link and resolve functions from these .o generated from GNU AS)
4. run `win64/m4as_all.bat` in `lisp-kernel` dir
5. Open this vcxproj file in Visual Studio, build in "Debug" and "x64"
6. Will get a `lisp-kernel.exe` in `lisp-kernel/x64/Debug/`

Some findings and potentially problem when building:
1. MSVC compiler fall behinds with gcc a lot. It's still ANSI C, doesn't support inline asm in 64 bit, and can't deal with very complex macros. Like the one I changed in x86-exceptions.c. Not sure it has some different macro expansions than gcc in other place that not raise syntax error.
2. Some unix header file doesn't exist, borrow from internet, mingw or write myself.
3. After all changes also try to rebuild in msys2, and it doesn't break.
4. Port it to Xcode maybe much easier because clang is modern and mac is posix


xrme:
Thank you for looking at this, but as #59 explains, we're going to give up on the idea to use Microsoft tools to build the lisp kernel.

PR #141, opened 2018-07-15
Rs1.11.5

ralfstoye:


xrme:
The ucontext_t fix is clearly a good thing to merge to 1.11, but let me think about the pathname-match-p for a day or two.

PR #182, opened 2019-01-03
Proposal of a fix to issue #181, file-position returning wrong values

jlahd:


phoe:
Is there a test written for this bug? If not, we'd need to write a regression test to make sure that the issue is actually fixed.

phoe:
TODO: check if one of the test cases at https://gitlab.common-lisp.net/ansi-test/ansi-test/blob/master/streams/file-position.lsp actually already reproduce this error. That is, once they stop failing with `#<SIMPLE-ERROR #x302003966F2D> [Not an ivector subtag: 182]` due to #246.

PR #227, opened 2019-10-15
Pull contrib changesets forward; Err when null FTD during cross-compile-cl in lib/nfcomp.lisp 

spchamp:
## Main Pull: Recovering the contrib subdirectory to the visible source repository

It seems that in the original SVN repository for CCL,the contrib subdirectory may have been moved to a location from which it may have been unreachable when the CCL git repository was created. Subsequently, the changeset in which it was moved is available in the Git changelog history as < 0f8979ca36d496ae5b5fb4979912ad3f18b7f629 > Of course, it appears in the Git changelog history as a deletion of everything under contrib. 

The original contrib directory can be retrieved from the changelog history, using git-subtree(1)

I'd used the following shell command to create the changesets denoted here as `[split: contrib]`
~~~~
/usr/bin/git subtree split --annotate "[split: contrib] " --rejoin -b ccl-contrib-subtree --prefix=contrib 2f20e80d1db7a5c030b21904fce6a1bb0f15f1ba
~~~~
This created a new series of changesets, each in effect representing one of the original commits but pulled forward as git-subtree synthetic changesets. Each of these changesets has been annotated as `[split: contrib]` in the resulting changelog history. 

This git call also created a local branch, `ccl-contrib-subtree` whose contents may seem to represent exclusively the contents of the synthetic `contrib` history, and the corresponding `contrib` subdirectory.

With the `--rejoin` option this branch was partially added back into the main series of changesets. 

Subsequently, this subtree branch and the corresponding contrib directory contents have been added back into the main source repository, with `git subtree add`
~~~~
/usr/bin/git subtree add --prefix contrib fbde32981ce80404c608c29c35bd0f8261eef3b9
~~~~

In effect, this may serve to pull the contents of the contib subdirectory forward -- perhaps, like something of a batch git cherry-pick.

The changeset as denoted in the initial `git subree split` call, i.e < 2f20e80d1db7a5c030b21904fce6a1bb0f15f1ba > was the second-to-last changeset under the original CCL contrib subdirectory. In effect, it was the last changeset before the contrib subdirectory became unreachable in the later Git changelog history.

So, in effect, this may serve to make the contents of the contrib subdirectory visible in the main repository, again.

## Second Changeset: Err when null FTD during cross-compile-cl in lib/nfcomp.lisp

The changesets created with the previous series of Git commands were added to the local repository, after I'd created the changeset < 931e7368ac153fe0fcd7bb3865d969d773e14836 > locally. That changeset may also be visible in this pull request. 

In that changeset, I'd added an error that may be reached in `%with-cross-compilation-target` when the `backend-target-foreign-type-data` accessor for the named `backened`  returns a null value. This may be reached during `cross-compile-cl` for instance, such as when the foreign type data has not been initialized to the target-backend. 

Candidly, I'd encountered such a situation, when endeavoring to cross-compile an armcl.image for an armcl built under a Termux environment, but then using a 32 bit CCL on an amd64 architecture (Debian 10)  

To that point, I'd managed to load the arm-arch.lisp and arm-backend.lisp files on the build host, though that was obviously not enough for producing an armcl.image.

Hopefully, the error message may serve to explain the situation to anyone else who might somehow reach that code path, in a build.


xrme:
First of all, thanks for being interested enough in CCL to make a pull request.

I don't think I like the idea of resurrecting `ccl:contrib;`.  Most of what is found there is unfortunately obsolete, particularly the Cocoa UI related items, and I thought it would be best just to leave it all out of the main ccl repository when I converted it to git (though as you point out, it's still in the history).

If you think it's a good idea to make that stuff visible again (and maybe it is), then perhaps we could make a separate ccl-contrib repository for it?

As for your cross-compiler tweak, I'd be glad to merge that separately.  I am still pretty much a git caveman, so a separate pull request for that would be the easiest way for me to do that.

spchamp:
Apology on the belated reply. I'll cherry-pick 931e7368ac153fe0fcd7bb3865d969d773e14836 into a separate branch off the Clozure main - will create a branch specific for that singular change, to ensure it won't pull in any later changes on the local devo branches - and submit a new pull request for that change, along with a second one

PR #233, opened 2019-10-28
Remove `void lazarus()` from Lisp kernel

### LINKED ISSUE #24: on Windows, (quit) sometimes hangs
http://trac.clozure.com/ccl/ticket/1345
http://trac.clozure.com/ccl/ticket/1393
http://trac.clozure.com/ccl/ticket/1408
http://trac.clozure.com/ccl/ticket/1409

possibly related:
http://trac.clozure.com/ccl/ticket/1142

For some reason, (quit) sometimes hangs on Windows.  It seems to happen with both 32- and 64-bit versions of CCL, and on both 32- and 64-bit versions of Windows.

http://trac.clozure.com/ccl/ticket/1345 has the most details.

> I've noticed that I can avoid the problem quite reliably when I change the compatibility setting of the wx86cl64.exe binary to Windows7. (This setting can be changed in the properties dialogue for the file wx86cl64.exe)

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/42682851-on-windows-quit-sometimes-hangs?utm_campaign=plugin&utm_content=tracker%2F27935804&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F27935804&utm_medium=issues&utm_source=github).
</bountysource-plugin>

phoe:
Fixes #24.

Given its name, I can infer that `lazarus` is mean to resurrect a dying thread if it still has a TCR: if TCR exists, then it calls start_lisp; if it doesn't, then it does nothing. start_lisp, in turn, is an assembly function that has the following comment inside it:

>  This is called from C code when a thread (including the initial thread) starts execution.  (Historically, it also provided a primitive way of "resettting" a thread in the event of catastrophic failure, but this hasn't worked in a long time.)

I assume that lazarus() is the function mentioned here. It does seem to try and "reset" a thread in the event of it doing an exit call (lazarus is bound to atexit() calls in the main function).

If that is true, therefore I infer that a possible fix be to remove this function from the CCL codebase, along with its atexit() bindings. This should solve GitHub issue #24 where a call that then calls lazarus() sometimes deadlocks when `(ccl:quit)` is called.

### DIFF
--- a/lisp-kernel/pmcl-kernel.c
+++ b/lisp-kernel/pmcl-kernel.c
@@ -1696,27 +1696,7 @@ check_arm_cpu()
 #endif
   return win;
 }
-#endif  
-
-void
-lazarus()
-{
-  TCR *tcr = get_tcr(false);
-  if (tcr) {
-    /* Some threads may be dying; no threads should be created. */
-    LOCK(lisp_global(TCR_AREA_LOCK),tcr);
-    tcr->vs_area->active = tcr->vs_area->high - node_size;
-    tcr->save_vsp = (LispObj *)(tcr->vs_area->active);
-#ifndef ARM
-    tcr->ts_area->active = tcr->ts_area->high;
-    tcr->save_tsp = (LispObj *)(tcr->ts_area->active);
 #endif
-    tcr->catch_top = 0;
-    tcr->db_link = 0;
-    tcr->xframe = 0;
-    start_lisp(tcr, 0);
-  }
-}
 
 #ifdef LINUX
 #ifdef X8664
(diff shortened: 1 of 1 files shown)


xrme:
This might have worked once on PowerPC (although comments seem to cast doubt on even that).  In any case, it's hurting us now.  I'm merging it.

phoe:
Thank you! 

Please prepare new bootstrapping binaries for Windows that include this patch. I will want to switch to them on Travis as soon as they are available.

xrme:
I'll upload new Windows binaries in the next day or two.

phoe:
Here is a Lisp-related cat picture so this thread doesn't die yet.

avoGK2d_460s

PR #235, opened 2019-10-29
Fix various typos

rprimus:
A second pass may be needed to correct inadvertent indentation changes.


rprimus:
Tue Oct 29 19:18:42 GMT 2019

@xrme HI.  Commit https://github.com/Clozure/ccl/pull/235/commits/fffe60118f87fa4fc1f090be5fc22603c611533f is using the `ccldoc` minor mode as well as setting `lisp-indent-offset 2`.

The `ref` sections still do not match the original. 

Using a vanilla emacs (emacs -Q) and `ccldoc.el` loaded, I get:
```elisp
  (defsection "Memory-mapped Files"
    (para "In release 1.2 and later, {CCL}
      supports "
	  (ref (glossentry "memory-mapped file") "memory-mapped files") ". On operating systems that support memory-mapped
      files (including Mac OS X, Linux, and FreeBSD), the operating
      system can arrange for a range of virtual memory addresses to
      refer to the contents of an open file. As long as the file remains
      open, programs can read values from the file by reading addresses
      in the mapped range.")
    "Using memory-mapped files may in some cases be more
      efficient than reading the contents of a file into a data
      structure in memory.

```
 Is there something I'm missing?

Cheers

xrme:
I expect some of the indentation was just done by hand.  Do what you can to minimize whitespace-only changes, but if some can't be easily avoided, don't worry about it.

rprimus:
Tue Oct 29 22:10:58 GMT 2019

@xrme Thanks for clearing that up!
The remaining 15 files have had their indentation updated.  The settings used should minimise 
the whitespace-only changes.

Cheers.

xrme:
Thank you, @rprimus.

PR #259, opened 2019-11-26
General CCL maintenance stuff

phoe:
This is a series of commits that attempt to clean CCL codebase up a little, fix some upstream ANSI-TESTs, add Travis CI configuration, and add an introspection tweak to method combination objects.

Please review it thoroughly before merging - I have run upstream ANSI-TESTs on all of these commits to ensure that no breaking changes were introduced.

I have attempted to add informative messages to almost each commit that explain why and what was changed. Only some trivial removals are messageless.

If this MR is too big and multi-themed, please close it - I'll create smaller ones that are more monothematic.


phoe:
Honestly, I think it is ugly, stupid, and an oversight in the standard. It is obvious to me that `#<NULL>` is better than `#<NULL >` and `#<#x41300>` is better than `#< #x41300>`, even though the latter two are mandated by the standard.

The alternative is to annotate these parts of these ANSI-TESTs with `:ANSI-SPEC-PROBLEM :PRINT-UNREADABLE-OBJECT-LONE-UGLY-SPACE` and refuse to comply with this part of the spec. I would like to do that, though I think I kinda lack formal power to do so.

Should we?

Dnia 27 listopada 2019 18:39:34 CET, "R. Matthew Emerson" <notifications@github.com> napisał(a):
>xrme commented on this pull request.
>
>If you think this is important, I'll let it pass, but I don't like this
>kind of legalistic nit-picking.  We are making things uglier to satisfy
>a pathological test case that is looking for something that can't
>possibly have any effect on a valid program.
>
>
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/Clozure/ccl/pull/259#pullrequestreview-323846927

-- Wysłane za pomocą K-9 Mail.

phoe:
Closing this for now - I'll re-sort the patches and submit smaller PRs.

PR #263, opened 2019-12-06
Add support for MSVC

phoe:
While browsing the GitHub network for CCL, I found this commit from @denim2x. I think it would be worthy to review, test and merge it - posting it here so it is not forgotten in time

TODO: one of the files in the pull request might contain LGPL code as a part of the Windows `gettimeofday` implementation. Example possible replacement that is licensed less liberally: https://stackoverflow.com/questions/1676036/what-should-i-use-to-replace-gettimeofday-on-windows

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ lisp-kernel/darwinx8664/mach_exc_server.c
 lisp-kernel/darwinx8664/probes.h
 lisp-kernel/static-linuxppc/external-functions.h
 
+*.patch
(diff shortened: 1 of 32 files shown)


eugeneia:
I checked this out today and it needed some changes to build, however the resulting kernel crashes with ` VirtualProtect spjump: 0x1e7 Attempt to access invalid address.`

So... I think it’d need some work. I’d be interested to hear how far you got this @denim2x 👀

xrme:
Thank you everyone for taking a look, but as #59 explains, we're going to give up (at least for now) on trying to use Microsoft tools to build the lisp kernel.

PR #302, opened 2020-04-26
encode-ffi-field - syntax checking for field properties

spchamp:
This patch represents a single changeset, developed previously under a local development branch. It's submitted here, from a patch branch under the thinkum-contrib mirror at GitHub

The patch was tested, previously, in October of 2019, under a Debian 10 amd64 installation. I' ve tested the ffigen5 build under FreeBSD 12.1. 

Subsequent of a review of the latest port update for CCL, I'd like to create a ccl-devel port under FreeBSD ports and test out ffigen5 under the latter. Pursuant of any tractable success in this, will try to publish the ports patch under a new FreeBSD ports mirror and mention to the CCL mailing list.

I've tested this patch out, only under interactive I/O on the Debian 10 host.

Original changelog message:
lib/db-io.lisp: Provide some syntax checking in encode-ffi-field

Under some build configurations, after processing OS header files in
some host environments, ffigen5 may produce field information having
non-positive offset and/or width properties, in some specific fiehelds.

This changeset provides a rudimentary parse-time syntax check for such
data, implemented here in `encode-ffi-field'. On encountering a negative
offset or width for a field, the type check will signal a continuable
error of type `simple-type-definition-error'. The continuation form
would throw to the symbol, `ignore-type', with no values.

Furthermore, if encountering a zero width for a field, tentatively the
updated `encode-ffi-field' would produce a warning condition of type
`simple-type-definition-warning'

This also updates the following functions, to gracefully handle the
catch symbol `ignore-type', such that may be thrown from the updated
`encode-ffi-field'
* save-ffi-objc-message
* db-write-byte-list
* save-ffi-function
* save-ffi-typedef
* save-ffi-struct
* save-ffi-union
* save-ffi-transparent-union
* db-define-var
* save-ffi-objc-class

The following condition classes are added, each in a tentative
definition, such as to be used in this simple type checking for ffigen
data, in `encode-ffi-field'
* simple-type-definition-error
* simple-type-definition-warning

This patch was originally developed circa October, 2019

The patch has been tested, to a limited extent, under a Debian 10 amd64
installation.

(cherry picked from commit 333a1f8591167a16e56d262b59d26c8ca852a613)

### DIFF
--- a/lib/db-io.lisp
+++ b/lib/db-io.lisp
@@ -1186,8 +1186,35 @@ satisfy the optional predicate PREDICATE."
            (length (length string)))
       (cons length (map 'list #'char-code string)))))
 
+(define-condition simple-type-definition-error (simple-error)
+  ())
+
+(define-condition simple-type-definition-warning (simple-warning)
+  ())
+
 (defun encode-ffi-field (field)
   (destructuring-bind (name type offset width) field
+    (cond
+      ;; check for some possible errors in the *.ffi encoding
+      ((minusp offset)
+       (cerror "Ignore this type definition"
+               'simple-type-definition-error
+               :format-control "Negative offset ~S for field ~S"
+               :format-arguments (list offset name))
+       (throw  'ignore-type (values)))
+     ((minusp width)
+      (cerror "Ignore this type definition"
+               'simple-type-definition-error
+               :format-control "Negative width ~S for field ~S"
+               :format-arguments (list offset name))
+      (throw 'ignore-type (values)))
+     ((zerop width)
+      ;; The warning text may be verbosely descriptive, using
+      ;; the unabridged contents of the FIELD description
+      ;; rather than the field name
+      (warn 'simple-type-definition-warning
+            :format-control "Field has zero width: ~S"
+            :format-arguments (list field))))
   `(,@(encode-name name)
     ,@(encode-ffi-type type)
     ,@(encode-uint offset)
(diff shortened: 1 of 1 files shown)


lib/db-io.lisp
```diff
@@ -1498,54 +1526,62 @@ satisfy the optional predicate PREDICATE."
...
+                    (pref key :cdb-datum.size) (length foreign-name))
+              (cdbm-put cdbm key contents))))))))
 
 (defun save-ffi-function (cdbm fun)
-  (let* ((encoding (encode-ffi-function fun)))
-    (db-write-byte-list cdbm
-			(ffi-function-string fun)
-			encoding
-			t)))
+  (catch 'ignore-type
```
phoe, line 1547: This seems superfluous. `DB-WRITE-BYTE-LIST` already has a `catch 'ignore-type`; do you expect that `encode-ffi-function` is going to `throw` as well in some way?

lib/db-io.lisp
```diff
@@ -1498,54 +1526,62 @@ satisfy the optional predicate PREDICATE."
...
+			  (ffi-function-string fun)
+			  encoding
+			  t))))
 
 (defun save-ffi-typedef (cdbm def)
-  (db-write-byte-list cdbm
-                       (ffi-typedef-string def)
-                       (encode-ffi-type (ffi-typedef-type def))
-		       t))
+  (catch 'ignore-type
```
phoe, line 1555: Ditto.

Issue #236, opened 2019-10-29
Broken executables on debian sid/ubuntu bionic


phoe:
I get an exception while reloading the boot image upon trying to rebuild the core.

Bootstrapping binary: lx86cl from April 2019
Commit: current master
OS: Debian sid
`uname -a`: Linux phoeframe 4.19.0-5-amd64 #1 SMP Debian 4.19.37-5 (2019-06-19) x86_64 GNU/Linux

```
Unhandled exception 11 at 0x16623fc0, context->regs at #xffd566a8
Exception occurred while executing foreign code
received signal 11; faulting address: 0x16623fc0
address not mapped to object
Main thread pid 30851
Current Thread Context Record (tcr) = 0xf7bca700
Control (C) stack area:  low = 0xffbf3000, high = 0xffd56b34
Value (lisp) stack area: low = 0x679f3000, high = 0x67b3f000
Exception stack pointer = 0xffd56afc
%eax = 0x16623fc0
%ecx = 0xffd56b00
%edx = 0x00000004
%ebx = 0x5988ff00
%esp = 0xffd56afc
%ebp = 0xffd56b08
%esi = 0x0000000c
%edi = 0x6fbd40b6
%eip = 0x16623fc0
%eflags = 0x00010286

%cs = 0x0023
%ds = 0x002b
%ss = 0x002b
%es = 0x002b
%fs = 0x0007
%gs = 0x0063
Lisp memory areas:
                code                  low                 high
         dynamic (9)           0x6fca29c0           0x70cb0000
         dynamic (9)           0x6fbc0000           0x6fca29c0
         dynamic (9)           0x6fca29c0           0x6fca29c0
         dynamic (9)           0x6fca29c0           0x6fca29c0
          static (8)              0x12000              0x14000
  managed static (7)           0x6bbc0000           0x6bbc0000
        readonly (4)           0x67bc0000           0x6bbc0000
          tstack (3)           0x67990000           0x679f2000
          vstack (2)           0x679f3000           0x67b3f000
          cstack (1)           0xffbf3000           0xffd56b34
Lisp kernel vc revision: v1.12-dev.4-35-ge8bd0293
Can't find symbol.
current thread: tcr = 0xf7bca700, native thread ID = 0x7883, interrupts disabled

(#x67B3EFD4) #x6FBD413D : #<Function %MAKE-RWLOCK-PTR #x6FBD40B6> + 135
[...]

Issue #274, opened 2020-01-03
SIGSEGV after startup, Debian 10, 32 bit


avodonosov:
Soon after starting CCL on Debian 10, 32bit, I have the below error.

It may happen seconds after startup, or a minute, or 4 minutes. It seems happening in some background thread.

Reproduced for CCL 1.11.5, CCL 1.10.

Rebuild of kernel doesn't help.

In case fix not possible soon, is there any workaround I can use?

If necessary, I can provide ssh access to the server where it happens, your public key will be needed for that.

```
? Unhandled exception 11 at 0x47690f, context->regs at #xbfb0fc28
Exception occurred while executing foreign code
 at _SPffcall + 87
received signal 11; faulting address: (nil)
unexpected si_code value: 128
? for help
[20245] Clozure CL kernel debugger: bt
current thread: tcr = 0xb7c19700, native thread ID = 0x4f15, interrupts enabled


(#xB7B97F7C) #x10429F35 : #<Function %NANOSLEEP #x10429C66> + 719
(#xB7B97FBC) #x104374E5 : #<Function HOUSEKEEPING-LOOP #x10437346> + 415
(#xB7B97FDC) #x10437A2D : #<Function (:INTERNAL (TOPLEVEL-FUNCTION (LISP-DEVELOPMENT-SYSTEM T))) #x1043794E> + 223
[20245] Clozure CL kernel debugger: Current Thread Context Record (tcr) = 0xb7c19700
Control (C) stack area:  low = 0xbf9ad000, high = 0xbfb10104
Value (lisp) stack area: low = 0xb7a4c000, high = 0xb7b98000
Exception stack pointer = 0xbfb10078

```

avodonosov:
Signal 11 is SIGSEGV, so if I understand correctly, the thread was sleeping in `nanosleep` call and was interrupted by the segmentation fault. 

What causes the segmentation fault I don't know. You should not have any custom code running to trigger the error; just waiting in REPL is enough.

xrme:
@avodonosov did the fix for #306 solve this?

avodonosov:
On the server where my error happened, it was recently found that memory was reduced by mistake. After the memory was increased again the error doesn't happen anymore.

So I can not say if it was related to PIE.

PR #306, opened 2020-05-05
Disable PIE on linuxx8632

nickg:
Many Linux distributions now build position independent executables by
default to better support ASLR. This means the text section can be
loaded randomly anywhere in the lower 3G of virtual address space on
32-bit Linux. Many of these addresses cannot be represented as fixnums,
which causes `_SPffcall` to crash early in boot calling a bogus address.

Explicitly disable PIE when building the kernel to avoid this. Fixes
Issue #236.

This is only a partial solution to the problem, however, as shared
libraries may still be loaded at high addresses. See Issue #274 which I
think is a related issue.

### DIFF
--- a/lisp-kernel/linuxx8632/Makefile
+++ b/lisp-kernel/linuxx8632/Makefile
@@ -73,8 +73,13 @@ OSLIBS = -ldl -lm -lpthread -lrt
 LINK_SCRIPT = # ./elf_x86_32.x
 USE_LINK_SCRIPT = # -T $(LINK_SCRIPT)
 
+# Disable PIE: we need to ensure the kernel functions are mapped to low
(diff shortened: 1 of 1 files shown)


phoe:
Thanks for this patch!

nickg:
> 
> This is only a partial solution to the problem, however, as shared
> libraries may still be loaded at high addresses. See Issue #274 which I
> think is a related issue.

OK ignore this bit: I studied the code more and normal FFI pointers are MACPTRs which can encode the full 32-bit range. This only affects pointers from `%kernel-import` where they're encoded as fixnums. So just turning off PIE is fine.

xrme:
Fantastic. Thank you for tracking this one down.

PR #313, opened 2020-05-16
Pprint

svspire:
PPrinting a defclass form now prints more nicely: Each slot definition gets its own line.


svspire:
Mistake: Too many commits in this PR.

PR #321, opened 2020-06-03
make reserve_tls_slots() more lenient

### LINKED ISSUE #311: Windows 10 error
Hi,

Trying CCL and I get this weird error:
```
C:\Local\CCL\ccl-dev>wx86cl64.exe
unexpected TLS index value: wanted 16, got 17
```
I cannot run the binary.
This is with Microsoft Windows [version 10.0.19041.208]
Any idea what is going on here?
Thanks,

### LINKED ISSUE #331: CCL not picking up local systems under ~/common-lisp
This is more of an inquiry.
Apparently the version of asdf that is part of CCL should be recent enough.
But for some reason CCL doesn't pick up the projects I have under `~/common-lisp`.

According to the asdf docs this should work:
https://www.common-lisp.net/project/asdf/asdf.html#Configuring-ASDF-to-find-your-systems

The same works with ECL and SBCL. All have the same quicklisp configuration in rc files.

Am I missing something?

astrangeguy:
Fixes #311

### DIFF
--- a/lisp-kernel/windows-calls.c
+++ b/lisp-kernel/windows-calls.c
@@ -1035,32 +1035,33 @@ init_winsock()
 void
 reserve_tls_slots()
 {
-  unsigned int first_available, n, i;
+  unsigned int last_slot, i, cnt_pad;
+  unsigned int pad_slots[30];
+  
+  /* Since there is no way to reserve specific TLS indices, we reserve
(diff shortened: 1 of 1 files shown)


xrme:
Thanks.

We don't actually need to reserve any TLS slots on 64-bit Windows at all (since we reserve a GPR to point at the TCR), but this could still help the 32-bit Lisp.

PR #337, opened 2020-08-13
Add :directory parameter to ccl:run-program

jdz:


phoe:
Will this function work with multiple threads without race conditions? We are accessing CWD, which is a process-global singular resource, so multiple threads which launch programs simultaneously may collide with each other on attempting to set this.

jdz:
> Will this function work with multiple threads without race conditions?

The directory is changed in the child process after fork (right before executing the external program), and I'm pretty sure the issues you mention are not relevant.  Am I missing something?

phoe:
OK - if we have already forked, then we are mutating the new process's CWD, and so the old process's CWD is untouched. Thanks for clarification.

jdz:
One thing I'm wondering is whether the external command should be executed if `chdir` fails?  I think not, so more work needs to be done here.

xach:
Imagine the command is “rm”. Directory context is very important.

jdz:
My (not very original) plan is to create a pipe and communicate any failures to parent process through that.

PR #346, opened 2020-09-04
Fix mingw compilation

johnfredcee:
Msys2 Mingw with gcc 10.2.0 no longer has mingw_dosmaperr available, so try to emulate it's logic in dos_maperr. Not sure this is the right thing to do, but something should be done.


surabax:
I have verified the fix, CCL successfully compiles and runs ccl-tests with the freshly updated MSYS2 on Windows 10 64-bit.

phoe:
Thank you for verifying; merging.

PR #348, opened 2020-09-27
fix interpretation of defsetf lambda lists

enometh:
* lib/setf.lisp: (defsetf) use CCL::%DESTRUCTURE-LAMBDA-LIST instead
of CCL::RENAME-LAMBDA-VARS to come up with a suitable setf
expansion.

https://lists.clozure.com/pipermail/openmcl-devel/2020-September/012217.html

Currently CCL does not expand defsetf lambda lists of the form
(defsetf get-foo (&key (add1 1) (add2 (+ add1 2)))
	 (data)
 `(setq $foo (- ,data ,add1 ,add2)))
(get-setf-expansion '(get-foo))
;; => The value #:ADD1 is not of the expected type NUMBER.

WIP. This patch tries to rectify that. With this patch

(setf (get-foo) 10) ;; should return 6

The patch is for review and doesn't fix indentation and involved the
use of of an unhygenic EXPRESSION symbol. This is to facilitate easy
review.

### DIFF
--- a/lib/setf.lisp
+++ b/lib/setf.lisp
@@ -216,8 +216,9 @@
       (unless (verify-lambda-list lambda-list)
         (signal-program-error $XBadLambdaList lambda-list))
       (let* ((store-vars (cons store-var mv-store-vars)))
-        (multiple-value-bind (lambda-list lambda-temps lambda-vars)
-                             (rename-lambda-vars lambda-list)
+        (multiple-value-bind (bindings lambda-temps)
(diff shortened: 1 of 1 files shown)


enometh (commented):
This should merge/commit/pull-request not be here. I think I messed up when trying to use the API and there isn't  a way to remove it Please delete it if you can.

I did push an update to the fix-defsetf-lambda-list - which should have been sufficient. The update fixes a bug where the declarations were misplaced

PR #349, opened 2020-10-18
Update "block" to new "text-block"

rprimus:
Sun Oct 18 17:28:23 BST 2020

Attempting to build the documentation was failing with:
```lisp
? (defparameter *d* (ccldoc:load-document "ccl:doc;manual;ccl.ccldoc"))
> Error: Unknown form (BLOCK "Tip" (PARA #:|Users of 64-bit platforms may need to pass options such
>                as "-m64" to gcc, may need to give the output library a different
>                extension (such as ".so"), and may need to user slightly different
>                values for other options in order to create an equivalent test
>                library.|))
> While executing: (:INTERNAL CCLDOC::FORM-CLAUSE), in process listener(1).
> Type :GO to continue, :POP to abort, :R for a list of available restarts.
> If continued: Output error text into document and continue
> Type :? for other options.
```

Investigation revealed `block` was changed to `text-block`: https://github.com/Clozure/ccldoc/commit/7c5dcf1513a4e8826718a70df8cfb7415401900e

This PR updates the manual for the above.

### DIFF
--- a/doc/manual/ffi.ccldoc
+++ b/doc/manual/ffi.ccldoc
@@ -1213,7 +1213,7 @@ uc_uc_test(unsigned char data)
       gcc -dynamiclib -Wall -o libtypetest.dylib typetest.c \\
       -install_name ./libtypetest.dylib
     ")
-    (block "Tip"
+    (text-block "Tip"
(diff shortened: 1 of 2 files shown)


xrme:
Thank you.

Issue #350, opened 2020-10-31
The compiler claims to mismatch the arguments.


hyotang666:
The behavior that I got is below (the dribbled one).

```lisp
"dribble"
? (lisp-implementation-version)
"Version 1.12 (v1.12) LinuxX8664"
? (defclass fun ()
  ()
  (:metaclass c2mop:funcallable-standard-class))
#<FUNCALLABLE-STANDARD-CLASS FUN>
? 
(defmethod initialize-instance :after
           ((fun fun) &key function &allow-other-keys)
  (c2mop:set-funcallable-instance-function fun function))
#<STANDARD-METHOD INITIALIZE-INSTANCE :AFTER (FUN)>
? 
(setf (symbol-function 'test) (make-instance 'fun :function #'car))
#<FUN #x3020018014FF>

;; Works fine.
? 
(test '(1 2 3))
1

;; But ccl:arglist could not detect correct arglist.
? (ccl:arglist 'test)
NIL
:ANALYSIS

;; The compiler claims.
? (compile nil (lambda () (test '(1 2 3))))
;Compiler warnings :
;   In an anonymous lambda form at position 24: In the call to TEST with arguments ('(1
;                                                                                         2
;                                                                                         3)),
;     1 argument was provided, but at most 0 are accepted
;     by the current global definition of TEST
#<Anonymous Function #x30200183958F>
NIL
NIL
? (dribble)
```

This may be related to #43 

What I want to do is muffling the compiler warnings (of course I can do it with cl:muffle-warning though).

digikar99:
Has there been any update to this? I can reproduce the issue in CCL 1.12 / x86-64 with [adhoc-polymorphic-functions](https://github.com/digikar99/adhoc-polymorphic-functions/) which basically is using a custom function class.

The most I have been able to figure out so far is that the warning comes from `fcomp-named-function`.

PR #369, opened 2021-05-17
Fix https://github.com/Clozure/ccl/issues/350 : set lfun-bits inside set-funcallable-instance-function

digikar99:
Please review if the bits set in the commit are the relevant ones. 

- I can load [dense-arrays](https://github.com/digikar99/dense-arrays) without excessive notes. This uses [adhoc-polymorphic-functions](https://github.com/digikar99/adhoc-polymorphic-functions/) which in turn uses custom function class underneath.
- The example given in #350 itself works.

### DIFF
--- a/level-1/l1-clos.lisp
+++ b/level-1/l1-clos.lisp
@@ -1822,6 +1822,9 @@ changing its name to ~s may have serious consequences." class new))
     (error "~S is not a funcallable instance" funcallable-instance))
   (unless (functionp function)
     (error "~S is not a function" function))
+  (lfun-bits funcallable-instance
+             (logior (lfun-bits funcallable-instance)
+                     (logand (lfun-bits function) (1- (expt 2 16)))))
   (setf (%gf-dcode funcallable-instance) function))
 
 (defmethod reinitialize-instance ((slotd slot-definition) &key &allow-other-keys)


phoe:
What is the root cause of the error? Why does this particular case of bit-setting fix it?

digikar99:
The most I have been able to figure out so far is that the warning comes from `compile-named-function` (and `nx1-check-call-args`, which calls `innermost-lfun-bits-keyvect`, so I presumed the bits themselves are not being set anywhere.)

phoe:
Is the `logand` required? Are the bits allowed to be longer than 16 bits?

digikar99:
The `logand` is required, because at least the first few bits are relevant for the type. If those bits are set, the object no longer remains a `adhoc-polymorphic-functions:adhoc-polymorphic-function`.

1.

`(lfun-bits funcallable-instance (lfun-bits function))` definitely results in a type-error

2.

And the following works, but I'm not sure why it should work:
 
```lisp
(lfun-bits funcallable-instance
           (logior (lfun-bits funcallable-instance)
                   (lfun-bits function)))
```

3.

Change in required arguments seems to change the last 10 bits - or, no...? Something elaborate:

```lisp
;;; Ignoring the warnings; and using https://www.rapidtables.com/convert/number/decimal-to-binary.html with 2s complement
;;; Is there a CL tool for the same?
CCL> (lfun-bits (lambda (a)))
-528482048 ;=> 11100000100000000000000100000000
CCL> (lfun-bits (lambda (a b)))
-528481792 ;=> 11100000100000000000001000000000
CCL> (lfun-bits (lambda (a b c d)))
-528481280 ;=> 11100000100000000000010000000000
CCL> (lfun-bits (lambda (&key a)))
-528482302 ;=> 11100000100000000000000000000010
CCL> (lfun-bits (lambda (&key a b)))
-528482302 ;=> 11100000100000000000000000000010
CCL> (lfun-bits (lambda (&optional a)))
-528482300 ;=> 11100000100000000000000000000100
CCL> (lfun-bits (lambda (&optional a b)))
-528482296 ;=> 11100000100000000000000000001000
CCL> (lfun-bits (lambda (&optional a b c d)))
-528482288 ;=> 11100000100000000000000000010000
CCL> (lfun-bits (lambda (&rest a)))
-528449536 ;=> 11100000100000001000000000000000
CCL> (lfun-bits (lambda (a &rest a)))
-528449280 ;=> 11100000100000001000000100000000

**Comment by phoe** [comment]
> ;;; Is there a CL tool for the same?

Yes, `format ~b`:

```lisp
CL-USER> (format nil "~32,'0B" (mod -528482288 (ash 1 32)))
"11100000100000000000000000010000"
```

PR #373, opened 2021-05-24
Minor Readme Update

digikar99:
It feels a bit more intuitive for the build-from-source instructions to be in the README itself, so the PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,16 +2,52 @@
 
 This is the source code for Clozure CL.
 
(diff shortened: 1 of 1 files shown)


xrme:
Maybe it would be even better for the README should actually be shorter rather than longer.

I think the main purpose is to provide enough information on how to download and run CCL. I think I'll skip the explanation of how CCL is composed of three parts, etc., and point more clearly to the current release.

PR #388, opened 2021-08-27
Add missing restart-failure condition

yitzchak:
Looking back at the commit history I don't see that this condition was ever defined. Followed the convention used by ABCL, CMUCL, ECL and SBCL in making this condition a control-error subclass.


yitzchak:
Any chance on this? It is a simple fix for an apparent oversight.

PR #407, opened 2022-03-05
Fix comma bug

Devon7:
Fix bug where erroneous commas in #A, #C or #S literals
create bogus literals which cause errors (possibly much) later.

Before:	#C(0 ,1) >> Error: The value (#:|`,| . 1) is not of the expected type REAL.

After:	#C(0 ,1) >> Error: … Comma inside complex literal

[Oops, accidentally deleted identical pull request #392]

### DIFF
--- a/level-1/l1-reader.lisp
+++ b/level-1/l1-reader.lisp
@@ -2802,7 +2802,7 @@ initially NIL.")
  #'(lambda (stream char arg)
      (require-no-numarg char arg)
      (let ((*backquote-stack* (when *backquote-stack* "complex")))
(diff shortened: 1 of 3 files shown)


Devon7:
Per a discussion at https://bugs.launchpad.net/sbcl/+bug/309093
about this same bug (fix released 2010-03-28)
perhaps we should imitate their more detailed error message,
i.e., change `"Comma inside ~A literal"`
to `"Backquoted comma inside ~A literal (not list or vector)."`

xrme:
I finally see that if I actually read what you wrote, and look at the diff with "hide whitespace" on, that this is actually a pretty simple change.

PR #416, opened 2022-06-11
Fixes #415: provide implementation type in implementation version string

### LINKED ISSUE #415: --version identifies the version, but not the compiler
```sh
$ ./ccl/dx86cl64 --version
Version 1.12.1 (v1.12.1) DarwinX8664
```

I would expect it to identify itself. Something like:

```sh
$ ./ccl/dx86cl64 --version
Clozure Common Lisp (v1.12.1) DarwinX8664
```

jakku-kun:
This allows the version string to display the lisp implementation type.
```
 % ./dx86cl64 --version
Clozure Common Lisp Version 1.12.1 (v1.12.1-13-g0f6d38ec) DarwinX8664
```

### DIFF
--- a/level-1/l1-boot-1.lisp
+++ b/level-1/l1-boot-1.lisp
@@ -68,7 +68,8 @@
     (format nil "~a~a~d" (string-capitalize os) cpu bits)))
 
 (defun lisp-implementation-version ()
-  (%str-cat "Version " (format nil *openmcl-version* (platform-description))))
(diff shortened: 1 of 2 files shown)


jakku-kun:
I am temporarily closing this PR. My current approach breaks the consistency between `lisp-implementation-type` and `lisp-implementation-version`. I will update this PR and re-open it when I have time to send a better patch.

Issue #356, opened 2021-01-06
Is support for Apple Silicon coming soon?


sjatkins:
SBCL isn't there yet and CCL from a mac store failed on Big Sur even without M1 chip.   

Trying from App Store on M1 I see this in the console that comes up.

Unhandled exception 10 at 0x7fff2023abdd, context->regs at #x20446a9b0
Exception occurred while executing foreign code
 at objc_msgSend + 29
received signal 10; faulting address: 0x18
? for help
[2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: Current Thread Context Record (tcr) = 0x504080
Control (C) stack area:  low = 0x204208000, high = 0x20446b298
Value (lisp) stack area: low = 0x20c00000, high = 0x20e4c000
Exception stack pointer = 0x20446ae88
[2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: [2134] Clozure CL kernel debugger: 
 symbol name :

xrme:
We want to get it running on Apple Silicon, but there's no timeline at this point.

joubertnel:
@sjatkins FWIW, I am able to _run_ the pre-built CCL x86_86 binary on M1/Big Sur. 
Trying out some cursory lisp expressions works (I haven't tried anything extensive). 

[...]

PR #419, opened 2022-07-24
Add error-handling, whitespace-safety, and Rosetta support to scripts/ccl & scripts/ccl64

gwerbin:
I noticed that on my ARM Mac system, the `ccl64` script had a fall-through case, which led to un-set shell parameters and eventually a confusing error message.

This PR patches the `ccl` and `ccl64` scripts to avoid such errors.

I also hard-coded Rosetta support into the `case` tree by using the x86-64 settings when ARM64 is detected and the environment variable `CCL_USE_ROSETTA=1`. I thought it was safer to require "opt-in" than "opt-out" given that there are apparently some problems running CCL under Rosetta (e.g. https://github.com/Clozure/ccl/issues/356).

### DIFF
--- a/scripts/ccl
+++ b/scripts/ccl
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+set -eu
+
 # Change the definition of CCL_DEFAULT_DIRECTORY below to refer to
 # your Clozure CL installation directory.  The lisp will use this
 # environment variable to set up translations for the CCL: logical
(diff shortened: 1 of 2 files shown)


scripts/ccl
```diff
@@ -8,43 +10,45 @@
...
+if [ -z "${OPENMCL_KERNEL:-}" ]; then
+  case "$(uname -s)" in
+    Darwin) case "$(arch)" in
+      ppc*) OPENMCL_KERNEL=dppccl ;;
+      i386) OPENMCL_KERNEL=dx86cl ;;
+      *)
+        echo "Unsupported architecture"
+        exit 1 ;;
+      esac ;;
+    Linux)
```
galdor, line 32: There might be an indentation issue here (or GitHub is not displaying things correctly).

PR #430, opened 2023-01-05
Loop fix

### LINKED ISSUE #300: Loop for by should only allow a positive number
The following works in CCL but not in SBCL:

```lisp
(loop for i below 10 for j by 0 collect (list i j))
;=> on CCL: ((0 0) (1 0) (2 0) (3 0) (4 0) (5 0) (6 0) (7 0) (8 0) (9 0))
; SIMPLE-TYPE-ERROR on SBCL
```

On mentioning it over at SBCL Bugs mailing list, I learnt that this is against the standard: http://www.lispworks.com/documentation/lw51/CLHS/Body/06_abaa.htm and only positive numbers are allowed for the by clause. 

As of 1.11.5, CCL also allows negative numbers.

svspire:
Fixes #300 (the Matt Kaufmann bug portion of that issue).

Summary: 
The type-spec that ANSI allows for the loop variable in a for-as-arithmetic subclause should NOT be slavishly checked against the :from, :to, or :by (and related) values because:

- If those values are constants, the CCL compiler is already capable both of inferring their types and coercing them to the stated type of the loop variable for initialization and loop-end tests.
- If the stated type-spec of the loop variable should include a range, that range never applies to the :by value, so for the loop macroexpander to typecheck said range against the :by value is an error. (This is true regardless of whether the :by value is a constant or a variable.)
- For all values :from, :to, and :by (as well as :downfrom, :upfrom, :downto, etc.), when they are variables, other mechanisms already exist for the programmer to declare their types, such as a let binding outside the loop form. To have the type-spec of the loop variable applied to them automatically would override (and sometimes even conflict with) those other mechanisms.

I am confident that the developers of the ANSI standard never intended for the type-spec of the loop variable to be blindly applied to the other values in a for-as-arithmetic subclause by the loop macroexpander. This PR fixes that problem.

### DIFF
--- a/library/loop.lisp
+++ b/library/loop.lisp
@@ -1857,11 +1857,14 @@ collected result will be returned as the value of the LOOP."
 			    (loop-make-variable
 			      (loop-gentemp 'loop-limit-) form indexv-type))))
 	 (:by
-	   (multiple-value-setq (form stepby-constantp stepby)
-             ;;; the by variable must be a positive number (so not zero)
-             (loop-constant-fold-if-possible form `(and ,indexv-type (real (0))) t))
-	   (unless stepby-constantp
-	     (loop-make-variable (setq stepby (loop-gentemp 'loop-step-by-)) form `(and ,indexv-type (real (0))))))
+          (let ((by-type (if (listp indexv-type) ; has range limits
+                             (ccl::numeric-ctype-class (ccl::specifier-type indexv-type)) ; ignore any range limits from indexv-type here
+                             indexv-type)))
+            (multiple-value-setq (form stepby-constantp stepby)
+              ;;; the by variable must be a positive number (so not zero)
+              (loop-constant-fold-if-possible form `(and ,by-type (real (0))) t))
+            (unless stepby-constantp
+              (loop-make-variable (setq stepby (loop-gentemp 'loop-step-by-)) form `(and ,by-type (real (0)))))))
 	 (t (loop-error
 	      "~S invalid preposition in sequencing or sequence path.~@
 	       Invalid prepositions specified in iteration path descriptor or something?"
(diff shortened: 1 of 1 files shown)


xrme (approved):
I think this looks OK, but this is `loop` we're talking about.

Clearly, we shouldn't try to declare that the `by` value is a bounded type.

I wondered if there was some way to to turn the `indexv-type` type specifier into a ctype and then determine the numeric-ctype-class of it, and write a suitable type specifier based on that. I don't know if that's even a good idea.

Given your [comments](https://github.com/Clozure/ccl/pull/430#issue-1519796482) on the issue, I think your changes are good.

svspire:
> I wondered if there was some way to to turn the indexv-type type specifier into a ctype and then determine the numeric-ctype-class of it, and write a suitable type specifier based on that. I don't know if that's even a good idea.

There is a way:
 (ccl::numeric-ctype-class (ccl::specifier-type '(integer 10 20))) ;==> INTEGER
That was the first thing I tried and then I realized it was pointless. If the values are constants, the compiler is already smart enough to DTRT with them and I don't think there would be any speed advantage to the compiled code (although it might be possible to make the compiler a tiny bit smarter if :by is a constant fixnum. Haven't looked at the assembler output to see if it's already doing that).

If :by etc are variables, I think it's important to let the programmer declare their types elsewhere if he/she wishes to do so. (That would be where the programmer could explicitly declare a :by variable to be a fixnum, for example.)

Might be worth further investigation to see if there's any advantage to applying the numeric-ctype-class to :by iff it's not already a declared variable with a type, but now we're talking about potentially deep surgery. As you say, this is `loop`, and in order to just fix MK's bug I was aiming for a minimal footprint (for now).

Issue #425, opened 2022-11-14
Can't build a working 1.12.1 (or latest master) on Windows 10


kyanha:
Toolchain: msys2 via its installer, with the mingw-w64-x86_64-toolchain group installed.

I get a bunch of warnings about getpagesize, wopen, gettimeofday, _open_osfhandle, and debug_show_registers not being defined, a warning of a read of 4096 bytes from a region of size 1, and a note of source object spjump_start of size 1.

But that's not what concerns me the most. 

The output executable can't be executed. I presume it's from the trailing output of the following command (executed by make):
```
/mingw64/bin/x86_64-w64-mingw32-gcc -Wl,--image-base=0x10000 -Wl,-script=pei-x86-64.x -m64 -g    -o ../../wx86cl64.exe  pad.o x86-spjump64.o x86-spentry64.o x86-subprims64.o pmcl-kernel.o gc-common.o x86-gc.o bits.o  x86-exceptions.o x86-utils.o image.o thread_manager.o lisp-debug.o memory.o windows-calls.o x86-asmutils64.o  imports.o lispdcmd.o plprint.o plsym.o xlbt.o x86_print.o -lpsapi -lws2_32 -static -lpthread
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: ../../wx86cl64.exe:/4: section below image base
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: ../../wx86cl64.exe:/20: section below image base
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: ../../wx86cl64.exe:/36: section below image base
```

When I attempt to run the output wx86cl64.exe, under msys2 bash I get `bash: ./wx86cl64.exe: cannot execute binary file: Exec format error`; when I try to run it from cmd, I just get a blue system modal dialog that says "This app cannot run on your PC".

This happens with the distributed 1.12.1 .zip file, as well as the 'v1.12.1' tag and 'master' branch.

Is there anything I might do differently to perhaps get it working? I would like some of the fixes committed to master in my lisp kernel.

Thanks for your help!

(note: this is not the same error or output as #408, so I'm filing it as a separate issue.)

xrme:
[...]

PR #432, opened 2023-01-10
Changes to support compiling under UCRT64

kyanha:
I was having trouble getting CCL to build a functional binary under UCRT64. With the changes @bshetty suggested to pei-x86-64.x (modified for UCRT64 instead of Cygwin), the binary would actually start but would give a VirtualProtect error. With the changes he later suggested for the linker (now in Makefile), it now generates a binary that will run and pass all the same tests that the distributed Cygwin version does.

This is NOT A PERFECT PATCH. Per https://github.com/Clozure/ccl/issues/425#issuecomment-1369620284, the newest compilers under Cygwin still won't create functional binaries.

I also do not actually know anything about GNU ld linker scripts, so I don't know everything that it's doing, or if it's doing too much.

### DIFF
--- a/lisp-kernel/win64/Makefile
+++ b/lisp-kernel/win64/Makefile
@@ -124,7 +124,7 @@ OSLIBS = -lpsapi -lws2_32 -static -lpthread
 
 
 ../../wx86cl64.exe:	$(KSPOBJ) $(KERNELOBJ) $(DEBUGOBJ) Makefile pei-x86-64.x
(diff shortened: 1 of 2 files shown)


kyanha:
I also need to point out that the path I put in pei-x86-64.x is compatible with UCRT64, but not with cygwin. Feel free to change the linker paths as necessary.

bshetty:
I think the official docs point to using msys so this should be good.

Regards,
Bharat

xrme:
I used your pull request as a starting point to make the Windows lisp kernels build with the MSYS2 environment.

See https://github.com/Clozure/ccl/commit/a6aeda7d7a7766eb0c39699a861fee00928ca00c

Issue #364, opened 2021-04-29
simple-file-error is not give a :format-control :-(


informatimago:
In %err-disp-common in l1-error-signal.lisp L90
```
                 (simple-file-error (make-condition condition-name
                                             :pathname (car errargs)
                                             :error-type format-string
                                             :format-arguments (cdr errargs)))
```
but:
```
(define-condition simple-file-error (simple-condition file-error)
  ()
  (:report (lambda (c s)
	     (apply #'format s (slot-value c 'error-type) 
		    (file-error-pathname c)
		    (simple-condition-format-arguments c)))))
```
and:
```
(define-condition simple-condition (condition)
  ((format-control :initarg :format-control
                  :reader simple-condition-format-control)
   (format-arguments :initarg :format-arguments
                     :initform nil
                     :reader simple-condition-format-arguments))
  (:report (lambda (c stream)  ;; If this were a method, slot value might be faster someday.  Accessors always faster ?
                               ;; And of course it's terribly important that this be as fast as humanly possible...
	    ;Use accessors because they're documented and users can specialize them.
            (apply #'format stream (simple-condition-format-control c)
                   (simple-condition-format-arguments c)))))
```
simple-file-error is a simple-condition that has a format-control
slot, but this slot is not filled by %err-disp-common, therefore
handlers that try to use that slot to format the error fail.

Note that %err-disp-common has a format-string argument that is correctly filled; I'd suggest adding a line:
```
                                                      :format-control format-string
```
to the make-condition call.

informatimago:
Proposed patch:
```
[pjb@despina org.xquartz:0 ccl-git 37Gi]$ git diff -t -w -b
diff --git a/level-1/l1-error-signal.lisp b/level-1/l1-error-signal.lisp
index 18b2cd78..35c373e7 100644
--- a/level-1/l1-error-signal.lisp
+++ b/level-1/l1-error-signal.lisp
@@ -87,6 +87,7 @@
                    (simple-file-error (make-condition condition-name
                                                       :pathname (car errargs)
                                                       :error-type format-string
+                                                      :format-control format-string
                                                       :format-arguments (cdr errargs)))
                    (undefined-function (make-condition condition-name
                                                        :name (car errargs)))

[pjb@despina org.xquartz:0 ccl-git 37Gi]$ 
```

PR #442, opened 2023-02-21
store errno in SIMPLE-FILE-ERROR conditions

galdor:
CCL signals SIMPLE-FILE-ERROR conditions for all file-related errors, but they do not contain any information to programmatically distinguish what the error actually is.

The most basic use case is the ability to detect that opening a file failed because it did not exist. Without this patch, it is impossible to make the difference between a file-not-found error and, for example, an IO error.

I originally wanted to introduced a new `FILE-DOES-NOT-EXIST` condition inheriting from `SIMPLE-FILE-ERROR` as SBCL does with `SB-EXT:FILE-DOES-NOT-EXIST`, but this would require seriously reworking the whole `%ERR-DISP-COMMON` thing, and I honestly have no idea where to start. Storing errno into the file error makes sense. Even if it is very low level, it provides the necessary information so that a caller can decide how to handle it, without changing any existing API.


galdor:
Btw I just realized that `SIMPLE-FILE-ERROR` was not exported. Would it be ok to export it with `SIMPLE-FILE-ERROR-ERRNO`, and if yes where should the export be?

xrme:
https://github.com/Clozure/ccl/issues/364 may also be in this area

galdor:
Indeed, it seems useful. But there is not much I can do about it, I do not have write access ;)

xrme:
I hate to discourage this, but it feels to me that this (expose errno in simple-file-error) is not the right way to go.

You said that you really wanted a `file-does-not-exist` error, and I'd rather see us pursue that. I see that there is a lot of hair around `%err-disp` and friends, and I don't blame you for wanting to look the other way.

galdor:
That would be better indeed; but in that case, I would simply ignore `%ERR-DISP` and signal the right condition directly at the source. I suspect `%ERR-DISP`, which seems to be only used to dispatch to the "right" condition type, could be removed altogether.

PR #451, opened 2023-07-17
Patch #'find-definition-sources. Fixes #450.

### LINKED ISSUE #450: find-definition-sources can't find internal callers
In the IDE, if you do "Show Callers" (C-M-c), it will show a dialog box of callers of whatever function name the cursor is on. If one of those callers is :INTERNAL (i.e. a definition inside a labels, flet, or an anonymous lambda), double-clicking on that item will throw an error.

Test procedure:
Open CCL in the IDE.

[note: it might be necessary to set up the following variables before rebuilding CCL. I typically
have all these set to T:
```
(setf ccl::*fasl-save-definitions* t
        ccl::*fasl-save-doc-strings* t
        ccl::*fasl-save-local-symbols* t
        ccl::*save-definitions* t
        ccl::*save-doc-strings* t
        ccl::*save-local-symbols* t
        ccl::*save-source-locations* t)
```
]

In the listener, type `find-definition-sources`.
[Ironically this is the function I'm patching, but it's a perfect illustration of the problem.]

Put the cursor somewhere on the word `find-definition-sources` and hit `C-M-c`.

This should produce a dialog box with 3 elements, the first of which is
(:INTERNAL HEMLOCK::GET-SOURCE-ALIST HEMLOCK:EDIT-DEFINITION)

Double-click on that line.

This will throw an error in the listener:
```
> Error: No known definitions for (:INTERNAL HEMLOCK::GET-SOURCE-ALIST HEMLOCK:EDIT-DEFINITION)
> While executing: CCL::TOPLEVEL-EVAL, in process Listener(4).
> Type cmd-. to abort, cmd-\ for a list of available restarts.
> Type :? for other options.

```

svspire:
This seems to fix the issue but I'm not sure it's the best way. For example it might be better to modify the "Show Callers" dialog so that double-clicking on an item in it always sends a data item to #'find-definition-sources that it knows how to deal with -- that's a less kludgy version of what I'm doing here.

This patch stops the error and the source file opens in the editor, but it doesn't put the cursor on the actual line where the internal definition occurs. You still have to look for that inside the toplevel function where it appears. It would be nice to fix that, but I have a feeling it might require modifications to the source-notes mechanism.

### DIFF
--- a/lib/source-files.lisp
+++ b/lib/source-files.lisp
@@ -462,7 +462,11 @@ The returned list is guaranteed freshly consed (ie suitable for nconc'ing)."
              (if (and (find-class 'xref-entry nil)
                       (xref-entry-p name))
                (setq implicit-type (xref-entry-type name) implicit-name (xref-entry-full-name name))
-               (setq implicit-type t implicit-name name)))))
+               (setq implicit-type t implicit-name
+                     (if (and (consp name) ; ensure things work for functions internal to methods and functions
+                              (eql :INTERNAL (car name)))
+                         (car (last name)) ; handles anonymous :INTERNAL as well as named :INTERNAL (i.e. flet and labels)
+                         name))))))
         (setq implicit-dt-class (class-of (definition-type-instance implicit-type)))
         (with-lock-grabbed (*source-files-lock*)
           (loop for (nil . dt) in *definition-types*


xrme (approved):
looks OK to me

PR #452, opened 2023-07-26
Fix symbol macro handling in SETF

### LINKED ISSUE #448: SETF does not always respect symbol macros
Especially when symbol macros are supplied through augment environment instead of being processed directly, SETF fails to respect them.

```lisp
CL-USER> (let ((env (augment-environment 
                     nil
                     :symbol-macro (list (list 'elt '(aref a 0))))))
           (macroexpand-1 'elt env))
(AREF A 0)
T
CL-USER> (let ((env (augment-environment 
                     nil
                     :symbol-macro (list (list 'elt '(aref a 0))))))
           (macroexpand-1 '(setf elt 1) env))
(SETQ ELT 1)
T
```

The following works though:

```lisp
CL-USER> (let ((a (make-array 10)))
           (symbol-macrolet ((elt (aref a 0)))
             (setf elt 1))
           a)
#(1 0 0 0 0 0 0 0 0 0)
```

ivan4th:
SETF macro was ignoring symbol macros in the place specs, so in some cases (such as MACROEXPAND-1) these were not being handled correctly. Fixes #448.

Note that GETF-SETF-METHOD does resolve symbol macros, and SETF macro has diverged from it.

### DIFF
--- a/lib/macros.lisp
+++ b/lib/macros.lisp
@@ -557,6 +557,8 @@
            (let* ((form (car args)) 
                   (value (cadr args)))
              ;This must match get-setf-method .
+             (when (symbolp form)
+               (setq form (%symbol-macroexpand form env)))
              (cond ((atom form)
                     (progn
                       (unless (symbolp form)(signal-program-error $XNotSym form))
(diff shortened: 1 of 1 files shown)


xrme:
Thank you for the fix.

PR #456, opened 2023-09-08
fix bugs in ~e format directive

### LINKED ISSUE #394: Incorrect printing of floating-point number
I'm using ccl 1.12 LinuxX8664.

```
? (format t "~,v,,ve" 15 2 1d10)
10.00000000000000D+8
```
The printed result is off by a factor of 10; the exponent should be 9, not 8.

Presumably there's a round-off problem in figuring out the exponent to print.

This is a serious problem in printing that I found with maxima's engineering format that uses the scale factor parameter for `~e` to print floating point number in the form x*10^n where n is a multiple of 3.

LdBeth:
I don't understand what exactly I did but this PR should fix all three issues identified in #394 .


xrme:
Good on you for having the courage to look around in there!

I'll review and hopefully merge soon.

LdBeth:
Thanks, I find there is one more problem I need to fix before ready for review:

When the number to be printed is zero, an extra leading zero would be printed before the decimal point, this is undesirable.

eg. `(format t "~10,5,,2e" 0.0) ; "00.0000E-2"`

LdBeth:
I'm running https://gitlab.common-lisp.net/ansi-test/ansi-test/-/blob/7d95090df17cdde93fb3db5b188651504926c442/printer/format/format-e.lsp

And found another bug:

```
(format t "~,2e" (- 1/2)) ; 5.00E-1
```

LdBeth:
I have ran most of the tests from https://gitlab.common-lisp.net/ansi-test/ansi-test slightly adapted at https://gist.github.com/LdBeth/edb3d7b9e3218240615e6e33a02440d8

The only issue it reveals is when `w` is a relatively small value, for example:

```
(format nil "~5e" 1.2)
;; ccl: "1.0E+0"
;; sbcl: "1.e+0"
```

However, I think this is within the range of the behaviors the ANSI CL allows so I also decided not to change it.

LdBeth:
I have also identified a bug that the float is not rounded when `d` is specified but `w` is not:

```
CL-USER> (format t "~,5,,-3e" 1.2344)
0.00012344E+4
CL-USER> (format t "~10,5,,-3e" 1.2344)
0.00012E+4
```

I have revised my change to also handle this problem.

xrme:
Thank you for these bug fixes, @LdBeth.

PR #484, opened 2024-04-24
Update README.md - add a workflow status badge

gvanuxem:
- See https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge


xrme:
Thank you for pointing out this feature and making this PR.

I'm going to decline merging this change for now for two reasons:
 * I don't think that the workflows are sufficiently well-developed for this to be really helpful.
 * I feel that adding this badge (or potentially many additional badges for other workflows) to the README.md file will add noise and be distracting.

Issue #504, opened 2024-07-04
TCP sockets stop working after a few writes, "Bad file descriptor (error #9) during write"


SuperDisk:
I'm messing around with the echo server example from [`websocket-driver`](https://github.com/fukamachi/websocket-driver), and I found that it has issues with CCL but not with other implementations.

After writing to a socket a couple times, the connection would inexplicably close-- I [removed the handler](https://github.com/fukamachi/websocket-driver/blob/ff606c1ff373b6b07380978f848cf6b6aa192977/src/ws/client.lisp#L208) which suppresses the error, and found that it's producing this:

```
on #<BASIC-TCP-STREAM ISO-8859-1 (NIL/4) #x302001DAEFFD> : 
Bad file descriptor (error #9) during write
```

The first few writes to the WebSocket work, and then after about 4 or 5 writes, this happens. The error comes from here in the call to `force-output`:

https://github.com/Clozure/ccl/blob/4c6c0fc12b44575ac27cfa23172fcf70b5d38ff2/level-1/l1-streams.lisp#L5749

The call is here in `websocket-driver`: https://github.com/fukamachi/websocket-driver/blob/ff606c1ff373b6b07380978f848cf6b6aa192977/src/ws/client.lisp#L207

Any idea what it could be?

Also, I just wanted to express my appreciation for your hard work maintaining this. I've been daily driving CCL while building a game and it's been quite nice.

xrme:
What is the output of `(lisp-implementation-version)` on your system, please? Maybe I could try to duplicate your issue.

The `(NIL/4)` in the print-object output means that CCL can't tell what kind of thing that file descriptor 4 is, which is not surprising given the `EBADF` error code. Normally, you'd expect to see something like `#<BASIC-TCP-STREAM ISO-8859-1 (SOCKET/7) #x3020005BF03D>`.

I'm guessing wildly, but maybe you could `trace` some relevant close function to see if there's something closing the stream unexpectedly? Is there possibly a race condition where a thread closes the socket while some other thread is still trying to access it?

SuperDisk:
I ran this code:

```lisp
 (ql:quickload :websocket-driver-client)

 (progn
   (trace (:package :websocket-driver.ws.client))
   (trace (:package :websocket-driver.ws.base)))

 (defparameter *client* (wsd:make-client "ws://localhost:5000/echo"))
 (wsd:start-connection *client*)
```
which should trace everything that happens on the websocket side.

[...]

PR #505, opened 2024-07-08
In make-tcp-socket, close stream (not just fd) when connect fails

xrme:
In #504, we see a problem where a socket is getting unexpectedly closed.

See the comments in the issue for details.

It might make sense to rewrite make-tcp-stream so that it doesn't create a stream at all until we have successfully created and configured the OS-level socket.

### DIFF
--- a/library/sockets.lisp
+++ b/library/sockets.lisp
@@ -685,7 +685,19 @@ the socket is not connected."))
 					      :port remote-port
 					      :allow-other-keys t
 					      keys))))
-               (%socket-connect fd socket-address timeout-in-milliseconds)))
+               (handler-bind
+                   ((socket-creation-error
+                     (lambda (c)
+                       (declare (ignore c))
+                       ;; When connect fails, the fd is no longer usable,
+                       ;; and must be closed.  Because we've already made
+                       ;; a stream using this fd, close the fd by closing
+                       ;; the stream.
+                       (close socket)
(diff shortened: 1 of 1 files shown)


gmpalter:
Here's an alternative that doesn't involve pulling the logic of `%socket-connect` out into `make-tcp-socket`. It catches the error signaled by `%socket-connect`, closes the stream and resets `fd`, and then lets the signaling continue.

```
diff --git a/library/sockets.lisp b/library/sockets.lisp
index b653b203..cd4a545f 100644
--- a/library/sockets.lisp
+++ b/library/sockets.lisp
@@ -685,7 +685,18 @@ the socket is not connected."))
                                              :port remote-port
                                              :allow-other-keys t
                                              keys))))
-               (%socket-connect fd socket-address timeout-in-milliseconds)))
+               (handler-bind (((or socket-creation-error socket-error)
+                                (lambda (c)
+                                  (declare (ignore c))
+                                  ;; When connect fails, the fd is no longer usable,
+                                  ;; and must be closed.  Because we've already made
+                                  ;; a stream with this fd, close the fd by closing
+                                  ;; the stream.
+                                  (close socket)
+                                  ;; Don't try to close fd again in the unwind-protect
+                                  ;; cleanup form.
+                                  (setq fd -1))))
+                 (%socket-connect fd socket-address timeout-in-milliseconds))))
            (setq fd -1)
            socket))
     (unless (< fd 0)
```

PR #527, opened 2025-02-24
Fix access to iobuf internals on Windows

johnfredcee:
Compiling with the latest mingw-w64-ucrt_x86_64 this failed to compile as it tries to access the internals of the FILE structure which have been made opaque.

This is my suggested fix.


xrme:
I myself am not an expert on selecting the right thing to do on Windows, but I'm happy to trust your patch for this. Thank you.

PR #528, opened 2025-03-13
Support for musl libc under Linux

se-mz:
I hope the commit messages explain the exact changes well enough.

I'm not super happy with the solution for `mcheck.h`, but as I understand it, `start-mtrace` and `stop-mtrace` are intended to be called manually for debugging purposes anyway, so it's not a big deal if the functions exist but fail.


xrme:
Thank you for your patch.

I am not worried about the mcheck stuff; as you note, it's for unusual debugging situations only anyway.

PR #529, opened 2025-03-19
Use `%uvector-replace` in `replace` even when subseqs differ in length

se-mz:
This significantly speeds up code that relies on the way `replace` implicitly intersects sequence bounds. For example, the `replace` calls below take about 150ms & 3ms respectively before the patch; both take about 3ms afterwards.
```
(let ((buf  (make-array 100000000 :element-type '(unsigned-byte 8)))
      (buf2 (make-array 10000000  :element-type '(unsigned-byte 8) :initial-element 1)))
  (time (replace buf buf2))
  (time (replace buf buf2 :start1 0 :end1 (length buf2)))
  nil)
```
Originally found by Gilbert Baumann a while ago.

### DIFF
--- a/lib/sequences.lisp
+++ b/lib/sequences.lisp
@@ -326,13 +326,13 @@
   (setq source-end (check-sequence-bounds source-sequence source-start
                                           source-end))
   (locally (declare (fixnum target-start target-end source-start source-end))
-    (let* ((typecode (typecode source-sequence))
-           (n (- source-end source-start)))
-      (if (and (not (listp source-sequence))
-               (simple-array-p source-sequence)
-               (= (the fixnum (- target-end target-start)) n)
+    (let ((typecode (typecode source-sequence)))
+      (if (and (simple-array-p source-sequence)
                (= typecode (typecode target-sequence)))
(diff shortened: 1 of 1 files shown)


xrme:
In the case of something like
`(make-array 100000000 :element-type '(unsigned-byte 8))`, where the memory is implicitly zero-filled (in CCL, anyway), it is possible that the memory will not actually be "real" yet, and that therefore the make-array call will take very little time.  (The memory will be provided zero-filled on demand pages from the operating system when we're consing into fresh memory.)

In a fresh lisp. you'll almost certainly see a time difference between `(make-array 100000000 :element-type '(unsigned-byte 8))` and `(make-array 100000000 :element-type '(unsigned-byte 8) :initial-element 1)`.

Your improvement to `replace` is probably a win anyway, but I wanted you to be aware of a possible source of timing error because you might accidentally be measuring the time needed for the operating system to provide zero-filled-on-demand pages for `buf` in the first `replace` form.

se-mz:
Good catch, but it seems like that doesn't matter here; I get comparable readings when initializing both with non-zero values.

PR #536, opened 2025-07-13
Use <= for three argument ~^

kg583:
The [HyperSpec Section  on the Escape Upward directive](https://www.lispworks.com/documentation/HyperSpec/Body/22_cib.htm) (`~^`) states the following.
> If three parameters are given, termination occurs if the first is less than or equal to the second and the second is less than or equal to the third.

However, the current implementation (inconsistently) uses strict inequality for this check.


xrme:
Thank you (very belatedly) for the patch.

rpitasky:
Thank you, we use tens of thousands of these in our code.

PR #541, opened 2025-08-12
Rebind the NFP depth in X862-LET and ARM2-LAMBDA-BIND

se-mz:
Fixes an oversight in cb1dded and 62c5caf respectively. The following code
returns 2000.0 instead of 1255.0 on x86-64 if the binding is missing:

```
(funcall (lambda ()
           (+ 255e0 (the single-float
                         (let ((foo (the single-float (+ 1000e0 (random 1))))
                               (bar 0))
                           (declare (ignore bar))
                           foo)))))
```

I couldn't find an example for the ARM backend since it rebinds the NFP depth
much more aggressively, but if rebinding is required for `lambda-bind` on x86, it
stands to reason that it's required on ARM as well.


se-mz:
Idle thought: this bug would have been caught if `x862-form` verified that it kept the stack depth intact (which should always be the case afaiu). Would it be worthwhile to add this as a sanity check? It'd probably make sense to do it for all stacks though, so maybe it's too heavy for something that touches every form.

Issue #520, opened 2024-10-02
Unhandled exception when running lparallel tests (1.13 LinuxARM32)


riktam:
Hi,
CCL either locks-up or crashes with the errors below when trying to run lparallel tests.

`Clozure Common Lisp Version 1.13 (v1.13) LinuxARM32

For more information about CCL, please see https://ccl.clozure.com.

CCL is free software.
It is distributed under the terms of the Apache License, Version 2.0.
? (ql:quickload :lparallel-test)
(lparallel-test:execute)To load "lparallel-test":
  Load 1 ASDF system:
    lparallel-test
; Loading "lparallel-test"

(:LPARALLEL-TEST)
? LPARALLEL-TEST::FORCE-CHAIN-TEST..........
LPARALLEL-TEST::END-KERNEL-TESTUnhandled exception 4 at 0x14a4a744, context->regs at #xa4e632d8
? for help
[3708] Clozure CL kernel debugger: :backtrace
[3708] Clozure CL kernel debugger: current thread: tcr = 0x1ad7480, native thread ID = 0xf2f, interrupts enabled

(#xA4E635D0) #x14A4A700 : #<Function CALL-WITH-WORKER-CONTEXT #x14a4a7de> + 60
(#xA4E635E0) #x00018C0C : (subprimitive ret1valn)
(#xA4E635F0) #x14A49494 : #<Function (:INTERNAL MAKE-WORKER-THREAD) #x14a494be> + 120
(#xA4E63620) #x14A49440 : #<Function (:INTERNAL MAKE-WORKER-THREAD) #x14a494be> + 36
(#xA4E63630) #x00018C0C : (subprimitive ret1valn)
(#xA4E63640) #x14991720 : #<Function (:INTERNAL %BINDING-DEFAULT-SPECIALS-WRAPPER BINDING-DEFAULT-SPECIALS) #x14991746> + 612
(#xA4E63670) #x14991704 : #<Function (:INTERNAL %BINDING-DEFAULT-SPECIALS-WRAPPER BINDING-DEFAULT-SPECIALS) #x14991746> + 584
(#xA4E636B0) #x00018C0C : (subprimitive ret1valn)
(#xA4E636C0) #x103CED28 : #<Function RUN-PROCESS-INITIAL-FORM #x14179f86> + 412
(#xA4E63700) #x103CECB8 : #<Function RUN-PROCESS-INITIAL-FORM #x14179f86> + 300
[...]

Issue #428, opened 2022-12-17
'Waiting for recursive lock' crash (CCL1.11, CCL1.12 on ARM32)


zinadv:
Copy this code to /tmp/bug.lisp.

```
(in-package :cl-user)

(ql:quickload :bordeaux-threads)
(shadow 'terminate)
(ql:quickload :iterate)
(use-package :iterate)

(defun worker-read-byte (&aux file byte)
  ;; sharing can be (:private :lock :external)
  (setf file (open "/tmp/bug.lisp" :element-type '(unsigned-byte 8)))
  (setf byte (read-byte file))
  (close file)
  byte)

(defun main (n)
  (let (threads)
    (setf threads
          (iter
            (for i :from 1 :to n)
            (for thread = (bt:make-thread #'(lambda () (worker-read-byte)) :name "tracking"))
            (collect thread)))
    (write
     (iter
       (for thread :in threads)
       (collect (bt:join-thread thread))))
    (terpri)))

(iter
  (main 16)
  (sleep 1))
```

After several iterations emits:

```
Unhandled exception 4 at 0x105901f8, context->regs at #x75fe3838
? for help
[30173] Clozure CL kernel debugger:
```

Works fine on amd64 (but I need ARM32). If remove work with files then works fine, but I need a lot of streaming.

Didn't use WITH-OPEN-FILE to check if it is argument :abort of function close. It can be NIL.

Tried every of (:private :lock (:external nil)) values to argument :sharing of function OPEN, because sometimes (with/out explicit JOIN-PROCESS) condition arises:

```
> Error: Current process #<PROCESS tracking(335) [Active] #x14E7FAF6> does not own lock #<RECURSIVE-LOCK [ptr @ #x12F7E20] #x140CF4D6>
> While executing: CCL::%UNLOCK-RECURSIVE-LOCK-OBJECT, in process tracking(335).
```

With explicit JOIN-PROCESS I see in logs (WRITE THREAD) that last thread is waiting for recursive lock too long.

xrme:
This sounds like a bug that we've struggled with on 32-bit ARM for a long time.

xrme:
[from the mailing list]

A test case is here:   https://github.com/jetmonk/openmcl-thread-test

Running (threadtest2)  triggers a modified with-lock macro that grabs a lock, checks that the current process in fact owns it, runs body, and again checks for ownership before releasing.    After many iterations, an error is thrown when the macro discovers that some other process stole the lock.

[...]

PR #555, opened 2026-06-23
Fix 32 bit ARM crashing bugs

varjagg:
This is to address #520 and #428 crashing bugs on 32 bit ARM. The patch was generated after some time with Codex and only approximate understanding on my part so discretion is advised. Nonetheless it appears to have stopped crashes on my test board both with our codebase and lparallel as referred in #520. Both were reproducing robustly before the change.

What it's supposed to do:

 - recognize the short ARM allocation prologue
  - computes how much of the allocptr adjustment has already happened
  - account the original free pointer
  - sets allocptr to the fully adjusted tagged value
  - moves PC to the allocation trap

### DIFF
--- a/lisp-kernel/arm-exceptions.c
+++ b/lisp-kernel/arm-exceptions.c
@@ -1752,6 +1752,24 @@ pc_luser_xp(ExceptionInformation *xp, TCR *tcr, signed_natural *alloc_disp)
     return;
   }
 
+  if (allocptr_tag == tag_fixnum) {
+    /*
+     * If a thread is suspended just before the instruction that adjusts
+     * allocptr, the valid allocptr value is still untagged.  Normalizing the
+     * thread for GC will replace allocptr with VOID_ALLOCPTR; leave the PC at
+     * the allocation start so the resumed thread takes the normal allocation
+     * trap and receives a fresh segment.
+     */
+    if (IS_SUB_RM_FROM_ALLOCPTR(instr) ||
+        IS_SUB_LO_FROM_ALLOCPTR(instr) ||
+        IS_SUB_FROM_ALLOCPTR(instr)) {
+      if (!alloc_disp) {
+        update_bytes_allocated(tcr, (void *)ptr_from_lispobj(cur_allocptr));
+        xpGPR(xp, allocptr) = VOID_ALLOCPTR;
+      }
+      return;
+    }
+  }
 
   
   if (allocptr_tag != tag_fixnum) {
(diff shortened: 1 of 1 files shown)


xrme:
"Approximate understanding" is what I have as well...

lisp-kernel/arm-exceptions.c
```diff
@@ -1671,6 +1671,92 @@ restart_allocation(ExceptionInformation *xp)
...
+  if (IS_SUB_LO_FROM_ALLOCPTR(instr)) {
+    return instr & 0xff;
+  }
+
+  if (IS_SUB_FROM_ALLOCPTR(instr)) {
+    return ror(instr & 0xff, (instr & 0xf00) >> 7);
+  }
+
+  return 0;
+}
```
xrme, line 1690: This looks like it was borrowed from the `allocptr_displacement()` function.

varjagg:
> "Approximate understanding" is what I have as well...

Right I really struggled with understanding the mechanism here, my only understanding is that it's a very specific memory/tagging race with exception handling.

If it's hard to make sense I suggest we wait a couple weeks while my boards run on this branch a bit. That doesn't add much on top of the years these bugs are there. If they still work then, merging perhaps still would make sense as in my experience arm32 runtime is effectively broken.

xrme:
OK, we'll wait while you run this code for a while.

I agree that the arm32 runtime is buggy, and given that your patch seems to help, I'm inclined to merge even if I don't completely understand it. 

The whole pc_luser_xp() scheme is extremely complicated. I'm becoming convinced that we'd be better off if we polled for gc/interrupts instead.

xrme:
Have you had the chance to stress-test your proposed change on your workload?

How are things looking?

varjagg:
There was a hangup on one of devices some 10 days ago but it was restarted before I could get to it to examine it. Still waiting if it reproduces, I say it's best we wait a bit.

varjagg:
It appears to work so far, with the caveat that I only tested it on Freescale i.MX6. But am still uneasy about this whole alien artifact nature of the patch.

xrme:
@varjagg I just committed a bunch of ARM-specific changes that insert memory barriers in places that I think the architecture's memory model requires.

If you get the time, please build from the master branch (maybe including your fix in this PR) and try your workload.  If anything good or bad happens, I'd be interested to hear about it.

xrme:
53a509a6cf attempts to fix a bug that could cause deadlock. Probably not related to this, since allocptr isn't involved.

Issue #85, opened 2017-11-20
lx86cl crashes on Ubuntu 17.10


xrme:
The lx86cl binary (both the one in the 1.11 release archive and a locally-built one) crash on Ubuntu 17.10.  The same problem seems to exist with the master branch.

Linux ubuntu 4.13.0-16-generic #19-Ubuntu SMP Wed Oct 11 18:35:14 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

dmesg output shows the following message:
[140567.196312] lx86cl[14478] bad frame in 32bit rt sigreturn frame:00000000ffa7ccc4 ip:6864eaae sp:67b6efec orax:ffffffffffffffff in lx86cl.image[67bf0000+ca3000]

xrme:
This happens also when using a 32-bit kernel.

Linux ubuntu 4.13.0-16-generic #19-Ubuntu SMP Wed Oct 11 18:33:49 UTC 2017 i686 i686 i686 GNU/Linux

[  880.620364] lx86cl[9010] bad frame in rt_sigreturn frame:bfb000f4 ip:10a5eaae sp:b7bf8fec orax:ffffffff in lx86cl.image[10000000+ca3000]

avodonosov:
Same problem when running CCL in docker image by CL Foundation, as github actions for testing.

Fragment from the host VM /var/log/syslog:
```
Sep 12 18:37:13 fv-az203-883 kernel: [  405.770120] lx86cl64[3797] bad frame in rt_sigreturn frame:0000000061352470 ip:300000c4815d sp:7f23be069fd8 orax:ffffffffffffffff in lx86cl64.image[300000000000+efd000]
```

avodonosov:
(As a workaround for Github Actions, I retry tests on CCL in case of crash. Implemented using the [[nick-fields/retry](https://github.com/nick-fields/retry)  custom step](https://github.com/nick-fields/retry) - see the above commit)

gornskew:
We have a reliable reproduction of this crash — roughly 1 in 10 process starts — in a publicly pullable Docker image, and some data that may help narrow it down.

Environment

CCL 1.13 (v1.13-21-gda7138ef) LinuxX8664, running as the gdl-ccl executable inside genworks/gendl:devo-ccl (Debian-based, public on Docker Hub)
Host: kernel 7.0.0-28-generic, Intel i7-12700K, Docker with default seccomp profile

Reproduction

for i in $(seq 1 60); do
[...]

PR #556, opened 2026-07-14
Preserve Linux xstate in exception contexts

luciusmagn:
Hi guys,

this is my first time contributing to Clozure, so I am not sure if there is a process I am forgetting.

I am submitting a (maybe a little opinionated) small fix for xstate handling.

On x86-64, the interrupted CPU register is carried by signals in a signal frame, and modern CPUs have AVX/other extended register state stored in the XSAVE area (see https://criu.org/Xsave, https://www.kernel.org/doc/html/latest/arch/x86/xstate.html, https://github.com/torvalds/linux/blob/master/arch/x86/include/uapi/asm/sigcontext.h).

Atm, when relocating a signal context, CCL copies only the old fixed-size FP/SSE portion of the frame. The copy can still advertise an extended XSAVE area that it does not contain, causing AVX and other extended register state to be discarded on signal return.

This should not break anything, if the extended metadata is missing or invalid, we copy the fp/sse portion and clear the marker so linux treats it as legacy.

I believe this is tangentially related to #85 which shows a bad frame crash as well. This should probably be fixed on x86-32 as well, I can do it if yall want.

Thank you for your time

### DIFF
--- a/lisp-kernel/x86-exceptions.c
+++ b/lisp-kernel/x86-exceptions.c
@@ -1522,6 +1522,47 @@ signal_handler(int signum, siginfo_t *info, ExceptionInformation  *context
 /* type of pointer to saved fp state */
 #ifdef X8664
 typedef fpregset_t FPREGS;
+
+#define CCL_FP_XSTATE_MAGIC1 0x46505853U
+#define CCL_FP_XSTATE_MAGIC2 0x46505845U
+#define CCL_MAX_XSTATE_SIZE (1024U * 1024U)
+#define CCL_FP_XSTATE_SW_BYTES_OFFSET 464U
+#define CCL_FP_XSTATE_EXTENDED_SIZE_OFFSET 468U
+#define CCL_FP_XSTATE_XSTATE_SIZE_OFFSET 480U
+
+static size_t
+linux_fpregs_size_in_bytes(FPREGS state, Boolean *extended_p)
+{
+  const BytePtr bytes = (BytePtr)state;
+  uint32_t magic1, extended_size, xstate_size;
+  uint32_t magic2;
+
+  memcpy(&magic1,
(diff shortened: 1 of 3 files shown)


luciusmagn:
I made a small example here: https://github.com/luciusmagn/xsave-failure

lisp-kernel/x86-exceptions.c
```diff
@@ -1522,6 +1522,47 @@ signal_handler(int signum, siginfo_t *info, ExceptionInformation  *context
 /* type of pointer to saved fp state */
 #ifdef X8664
 typedef fpregset_t FPREGS;
+
+#define CCL_FP_XSTATE_MAGIC1 0x46505853U
+#define CCL_FP_XSTATE_MAGIC2 0x46505845U
```
xrme, line 1541: I guess that you are defining these here because you don't want to pull in some header file.  A short comment about where the constants come from will probably help a future reader.

xrme (changes requested):
Please write a comment about the source of magic number words and the purpose of the other constants and then this will be ready to merge

A nice change. If you want to make a similar, separate change for 32-bit x86, I'll be happy to accept that.

lisp-kernel/x86-exceptions.c
luciusmagn: Yep! I have added one! :)

luciusmagn:
> Please write a comment about the source of magic number words and the purpose of the other constants and then this will be ready to merge
> 
> A nice change. If you want to make a similar, separate change for 32-bit x86, I'll be happy to accept that.

Thanks a lot, I have added support for 32-bit x86, please check it out whenever you can :)

xrme (approved):

xrme:
Thank you for the patch.

PR #560, opened 2026-07-27
arm64: interim port series from a parallel Linux/arm64 build (54 commits, based at 9c61574)

cyclistmass:
Interim patch series from a parallel Linux/arm64 port built on this branch. Based
at `9c61574` (the `arm64` head when this was opened), so every commit applies as-is.

Offered for cherry-picking rather than as a single take-it-or-leave-it merge —
each commit is independently scoped, has its own rationale in the message, and
cites the donor it was ported from (PPC64 in most cases, ARM32 or x8664 where
PPC has no analog). Nothing here depends on our own out-of-tree overlay: these
are all changes to files already in this branch.

### Where this gets to

With these applied plus our overlay, the arm64 build boots to a REPL, loads all
of level-1 and CLOS, runs its own resident compiler, does `COMPILE-FILE`, and
runs the ANSI test suite to completion: **21679 executed, 20 failures, no
exclusions** — 21659 passing, 99.908%. Every number quoted here is from a
measured run, not a projection.

### Rough map

42 files, +1094/-225.

- **`lisp-kernel/`** (10 files) — `.SPffcall` and the bind family bodies, image
  ABI version, NIL init, lisp-globals geometry, and `MIN_TSTACK_SIZE`.
- **`compiler/`** (10 files, 8 of them `ARM64/`) — UUO template renumbering to match
  `arm64-uuo.s`, register-bearing UUO templates, `arm64-subprimitive-offset`,
  `arm642-box-u32/-s32`, register maps sized for 32 GPRs rather than 16, NFP
  scratch-register fix, constant-index gating, the `DISASSEMBLE` entry point,
  and the compiler-correctness fixes found by the ANSI suite.
- **`level-0/`, `level-1/`, `lib/`** (21 files) — mostly missing `#+arm64-target`
  arms in shared code, where the absence silently yields NIL rather than
  failing loudly: `SYMBOL-NAME` returning its docstring, `SYMBOLP`,
  `SUBTAG-BYTES`, `IMMEDIATE-P`, `ARRAY-ELEMENT-TYPE` returning NIL for every
  array, the CLOS/dcode function builders, header-type tables, and the arch
  quads that load the arm64 support files.
- **`xdump/`, `lib/nfcomp.lisp`** — the fasl writer/scanner and cold-load
  function tagging for the arm64 tag scheme.

### Added since the PR was opened (commits 48-54)

- **48** — a UUO code for slot-unbound. This was the last `brk #xf0NN`
  placeholder in our tree; note `brk` does not trap as a UUO on arm64, so any
  remaining placeholder is a silent wrong-code path rather than an error.
- **49** — `MIN_TSTACK_SIZE` (1<<18) is *exactly* the tstack demand of reading a
  literal at `ARRAY-RANK-LIMIT`, so the suite died with zero headroom. Node size
  cancels out of the arithmetic, so every tsp-based port has the same
  zero-headroom collision, not just arm64.
- **50** — `%make-complex-{single,double}-float` kept ARM32's register-*pair*
  model instead of arm64's one-register-two-lane one.
- **51** — **removes `fulltag_function`**: a function becomes an ordinary
  miscobj (`fulltag_misc` + `subtag_function`), the code-vector offset goes
  −7 → −4, and `functionp` tests the subtag. This implements the design you
  described in the 2026-07-27 thread, and it dissolves the macro/special-operator
  fcell problem the original PR text listed as deliberately out of scope
  (`(funcall 'quote 1)` now signals rather than faulting). It is build-confirmed
  in our lane with zero regressions. If you would rather land your own version,
  say so and we will drop this commit and rebase onto yours — we are not asking
  you to take our shape over yours, only offering a working one.
- **52** — `ARM64-XDISASSEMBLE`. `compiler/ARM64/arm64-disassemble.lisp` supplies
  the whole disassembler but no CCL-facing entry point, so `DISASSEMBLE` had
  nothing to dispatch to.
- **53, 54** — **an upstream-wide defect, not an arm64-only one.** See below.

### Commits 53/54 are worth your eye even if you take nothing else

`arm642-elide-pushes` turns `(vpush P) … (vpop Q)` into a compensating copy
`Q := P`. When `P` is assigned inside the elided range the copy must be emitted
early — but the guard that permits the elision is the weaker condition *"Q's
last read precedes P's first write"* (`vinsn-sequence-refs-reg-p` returns the
last ref, `vinsn-sequence-sets-reg-p` the first set). That guard proves a safe
slot **exists** between those two points; the insertion ignored it and anchored
the copy at the push, clobbering a live base register ahead of its own uses.

Fixing that one anchor took our ANSI failures from **49 to 20**. The blast
radius is any three-operand form whose operands are nested accessors off a
common base — 29 tests, where we had predicted 9.

Reading across the other backends at `8b1ed247`:

| | vsp leg | nfp leg |
|---|---|---|
| `x862-elide-pushes` (x862.lisp:4054) | **already correct** — anchors at `popped-reg-is-reffed` (4164) | unguarded, anchors at the push (4093) |
| `arm2-elide-pushes` (arm2.lisp:3267) | computes `popped-reg-is-reffed` (3373) and then ignores it (3432) — the same defect | unguarded (3307) |
| `ppc2-elide-pushes` (ppc2.lisp:2794) | no vsp leg | unguarded (2817) |
| `arm642-elide-pushes` | the defect we hit | unguarded |

So x862's vsp leg is the correct donor and both ARM ports dropped the anchor
when adapting it. **`arm2.lisp:3432` looks like a live miscompile in the ARM32
backend**, reachable by the same source shapes — we have not built ARM32 to
confirm it, so treat that as a reading of the source rather than a measurement.

Commit 54 applies the same rule to the nfp leg, which had the wrong anchor *and*
no guard at all. It is offered on the identical-defect argument, not on a
failing test: its acceptance criterion was a full-suite run that does not move,
and that is what it did (21679/20, byte-identical failure list).

### Caveats, stated plainly

- Two hunks are tagged `ARM64-DEVIATION` in-file: places where AAPCS64 or the
  A64 encoding has no PPC-shaped equivalent and we made a judgement call. Those
  are the ones most worth your eye.
- Commit 9 renumbers the `arm64-asm` UUO templates to match the scheme in your
  `arm64-uuo.s`. It is mechanical, but it is a wide rename, so it is the most
  likely to conflict with in-flight work on your side — happy to rebase or drop it.
- A few kernel changes assume struct offsets and xtype values that we derived
  rather than found declared. Where we did that we said so in the commit
  message; if you would rather those live in `arm64-constants.h` we will move them.
- Commit 16 (`l1-clos`) is scoped to `level-1/l1-clos.lisp` only. Its
  `l1-clos-boot.lisp` half is already carried by commit 1.
- The series is verified to apply to a virgin checkout of this branch in order,
  **54/54**, with the `git am` result tree byte-identical to a `git apply`
  dry-run tree.
- The branch has moved on since `9c61574` (`1db0ab68` as of this update,
  including a large `arm642.lisp` expansion). We have not rebased yet; say the
  word and we will. `arm642-elide-pushes` still carries the commit-53 defect at
  `1db0ab68`.

Happy to split this into smaller PRs by area, reorder, or drop anything you
don't want.

### DIFF
--- a/compiler/ARM64/arm64-arch.lisp
+++ b/compiler/ARM64/arm64-arch.lisp
@@ -185,6 +185,8 @@
 (defconstant no-thread-local-binding-marker subtag-no-thread-local-binding)
 (define-subtag lisp-frame-marker fulltag-imm-1 5)
 (defconstant lisp-frame-marker subtag-lisp-frame-marker)
+(define-subtag stack-alloc-marker fulltag-imm-1 6)
+(defconstant stack-alloc-marker subtag-stack-alloc-marker)
 
 ;;; Extended type codes ("xtypes") for wrong-type UUOs.  The 8-bit
 ;;; expected-type field of a wrong-type UUO holds either a lisptag, a
--- a/compiler/ARM64/arm64-arch.lisp
+++ b/compiler/ARM64/arm64-arch.lisp
@@ -1290,3 +1290,102 @@
 (defparameter *image-abi-version* 1046)
 
 (provide "ARM64-ARCH")
+
+;;; ------------------------------------------------------------------
+;;; Subprimitives the linuxarm64 port adds to the table above.  This must
+;;; sit AFTER *arm64-target-arch*, because it refreshes the arch struct's
+;;; captured table as well as the variable: subprim-name->offset resolves
+;;; through the struct (compiler/subprims.lisp:53), so extending only the
+;;; variable gives ".SPRPLACD not found" while the variable plainly holds it.
+;;; ------------------------------------------------------------------
+;;; SPDX-License-Identifier: Apache-2.0
+;;;
+;;; PROPOSED (ratify with Matt): extend arm64-arch.lisp's *subprims*
+;;; vector (his :359-377 ends with ";; ..." -- knowingly incomplete)
+;;; with the subprimitives the level-0 drafts and our kernel spentry
+;;; drafts reference.  Offsets continue his vector's numbering (same
+;;; *subprims-base* / *subprims-shift* arithmetic); the ORDER below is
+;;; the proposal.  Kernel-side bodies for every entry exist in
+;;; upstream-port/lisp-kernel/spentry-{A..E}.s.
+;;;
+;;; Loaded by tools/upstream-assemble-test.lisp after Matt's LAP layer.
+
+
+(let ((offset (+ *subprims-base* (* (length *subprims*) (ash 1 *subprims-shift*))))
+      (step (ash 1 *subprims-shift*)))
+  (flet ((define-subprim (name)
+           (ccl::make-subprimitive-info :name (string name)
+                                        :offset (prog1 offset (incf offset step)))))
+    (macrolet ((defsubprim (name)
+                 `(define-subprim ',name)))
+      (setq *subprims*
+            (concatenate 'vector
+                         *subprims*
+                         (vector
+                          ;; multiple values (spentry-C)
+                          ;; boxing/unboxing (spentry-A)
+                          ;; special variables (spentry-A/C)
+                          ;; EGC-barrier stores (spentry-B)
+                          ;; arrays (spentry-B)
+                          ;; lexpr / rest args (spentry-D/E)
+                          ;; error signalling / dispatch (spentry-D)
+                          ;; FFI (spentry-E)
+                          (defsubprim .SPffcall-return-registers)
+                          ;; vinsn wave V1 demands (U4 in
+                          ;; arm64-vinsns-additions-w1-report.md);
+                          ;; lambda-binding subprims (spentry-D bodies)
+                          ;; call dispatch (spentry-D: funcall@145,
+                          ;; tfuncallgen@662, tfuncallslide@689) —
+                          ;; gate-26 funcall demand (w4 call cluster)
+                          (defsubprim .SPtfuncallvsp)
+                          ;; generic uvector element access (spentry-B:
+                          ;; misc_ref@588, misc_set@819) — w3a misc-ref/
+                          ;; misc-set vinsns resolve these by name (U2a)
+                          ;; gate-27 definitive referenced-vs-registered
+                          ;; scan: barrier'd cons/gvector mutation
+                          ;; (rplaca@D:181, rplacd@D:228, gvset@B:81)
+                          ;; and builtin tail (logior@D:1270, ash@D:1298)
+                          ;; make-closure cluster (gate-33): stack
+                          ;; gvector builder (spentry-B stkgvector@769)
+                          ;; %gvector handler (w5): heap gvector builder
+                          ;; (spentry-B gvector@564-584, verified body)
+                          ;; MV-pass + known-tail-call family (gate-34
+                          ;; sibling sweep; PPC64 vinsns 3947-3957;
+                          ;; spentry-D bodies 279/1581/720/751/772/780)
+                          ;; builtin-call dispatch (gate-35; spentry-D
+                          ;; bodies 1160/1174/1186/1198/1210)
+                          (defsubprim .SPcallbuiltin)
+                          (defsubprim .SPcallbuiltin0)
+                          (defsubprim .SPcallbuiltin1)
+                          (defsubprim .SPcallbuiltin2)
+                          (defsubprim .SPcallbuiltin3)
+                          ;; variable-size uvector alloc (gate-36;
+                          ;; spentry-A misc_alloc@450 / _init@679)
+                          ;; MV fit (gate-37; spentry-D fitvals@300)
+                          ;; checked special-ref (demand-scan wave;
+                          ;; spentry-A specrefcheck@372)
+                          ;; &optional defaulting (cut-3 wave;
+                          ;; spentry-D default_optional_args@357)
+                          ;; lexpr entry + MV slide (cut-5 wave;
+                          ;; spentry-E lexpr_entry@604, spentry-D
+                          ;; mvslide@1532)
+                          (defsubprim .SPlexpr-entry)
+                          ;; stack/list allocation family (gate-39
+                          ;; sibling sweep; PPC64 vinsns 3983-3999;
+                          ;; spentry-A 504/579/619/705/736,
+                          ;; spentry-B 371/386/449)
+                          ;; nth-value (w9; spentry-D nthvalue@317)
+                          ;; integer-sign (w9; spentry-A integer_sign@399)
+                          ;; catch/throw/unwind-protect/progv cluster (w10;
+                          ;; spentry-C-bind-catch-throw.s:319/326/334/1663/
+                          ;; 343/440/550, spentry-B:478/1044)
+                          (defsubprim .SPnmkunwind)))))))
+
+;;; The extension above rebinds the arm64::*subprims* VARIABLE, but
+;;; ccl::subprim-name->offset resolves through the arch STRUCT's
+;;; captured table (compiler/subprims.lisp:53 -> arch::target-subprims-
+;;; table of backend-target-arch), which *arm64-target-arch* filled at
+;;; its own load time (arm64-arch.lisp:987 :subprims-table) — gate-28:
+;;; ".SPRPLACD not found" while the variable held it.  Refresh the
+;;; struct so both lookup paths see the extended vector.
+(setf (arch::target-subprims-table *arm64-target-arch*) *subprims*)
--- a/compiler/ARM64/arm64-arch.lisp
+++ b/compiler/ARM64/arm64-arch.lisp
@@ -1268,6 +1268,8 @@
 (defconstant fasl-version #x1)
 (defconstant fasl-max-version #x1)
 (defconstant fasl-min-version #x1)
-(defparameter *image-abi-version* #x1)
+;; Next free after ARM32's 1045; the placeholder #x1 fails every
+;; kernel's "image too old" minimum check (image.h ABI_VERSION).
+(defparameter *image-abi-version* 1046)
 
 (provide "ARM64-ARCH")
--- a/compiler/ARM64/arm64-arch.lisp
+++ b/compiler/ARM64/arm64-arch.lisp
@@ -620,16 +620,22 @@
   flags
   link)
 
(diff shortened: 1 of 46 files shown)


cyclistmass:
Pushed seven more commits (47 → 54) and updated the description. Same base,
`9c61574`; the whole series still applies to a virgin checkout of this branch in
order, 54/54, and the `git am` tree is byte-identical to a `git apply` dry-run tree.

- **48** UUO code for slot-unbound (our last `brk` placeholder — worth noting
  `brk` does not trap as a UUO on arm64, so a placeholder is a silent wrong-code
  path, not an error)
- **49** `MIN_TSTACK_SIZE` is *exactly* the tstack demand of reading at
  `ARRAY-RANK-LIMIT`; node size cancels, so this is not arm64-specific
- **50** `%make-complex-{single,double}-float` was using ARM32's register-pair
  model rather than arm64's one-register-two-lane one
- **51** removes `fulltag_function` — a function becomes an ordinary miscobj,
  code-vector −7 → −4, `functionp` reads the subtag. This is the design from the
  2026-07-27 thread, build-confirmed here with zero regressions; it dissolves the
  macro/special-operator fcell problem the original PR text listed as out of
  scope. If you would rather land your own version, we will drop this and rebase
  onto yours.
- **52** `ARM64-XDISASSEMBLE` — the disassembler was all there, with no
  CCL-facing entry point for `DISASSEMBLE` to dispatch to
- **53, 54** the `elide-pushes` copy-placement fix, below

**53 is the one I would look at first, because it is not arm64-only.**
`arm642-elide-pushes` anchored the compensating copy at the push site when the
pushed register is reassigned in the elided range. The guard that permits the
elision is the weaker "popped's last read precedes pushed's first write", i.e. it
proves a safe slot *exists* — and the insertion ignored it, clobbering a live
base register ahead of its own uses. That one anchor was worth **49 → 20** ANSI
failures here; the blast radius is any three-operand form whose operands are
nested accessors off a common base, which was 29 tests where we had predicted 9.

Reading the other backends at `8b1ed247`:

| | vsp leg | nfp leg |
|---|---|---|
| `x862-elide-pushes` (x862.lisp:4054) | **correct** — anchors at `popped-reg-is-reffed` (4164) | unguarded, anchors at the push (4093) |
| `arm2-elide-pushes` (arm2.lisp:3267) | computes `popped-reg-is-reffed` (3373), then ignores it (3432) | unguarded (3307) |
| `ppc2-elide-pushes` (ppc2.lisp:2794) | no vsp leg | unguarded (2817) |

x862's vsp leg is the correct donor; both ARM ports dropped the anchor when
adapting it. **`arm2.lisp:3432` therefore looks like a live miscompile in the
ARM32 backend** — that is a reading of the source, not a measurement, since we
have not built ARM32. Commit 54 does the nfp leg on the identical-defect
argument only: its acceptance criterion was a full-suite run that does not move,
and it didn't (21679/20, byte-identical failure list).

Current state here: the suite runs to completion at **21679 executed, 20
failures, no exclusions**. Also noticed that the branch has moved past
`9c61574`; happy to rebase whenever it suits you. `arm642-elide-pushes` still
carries the commit-53 defect at `1db0ab68`.

cyclistmass:
Measured update, since both PRs quote an older number.

With this content applied, the ANSI suite now runs to completion at
**21675 of 21679 passing — 99.98%, four failures, zero exclusions**
(`:EXCLUDED NIL`, `:EXCLUSIONS-NOT-FOUND NIL`, `EXIT:0`). It was 20 failures
this morning.

The four:

```
PRINT.SHORT-FLOAT.RANDOM    PRINT.SINGLE-FLOAT.RANDOM
COMPILE-FILE.2              COMPILE-FILE.2A
```

**`COMPILE-FILE.2` and `.2A` fail identically on x86-64 CCL 1.12.2**, with a
byte-identical `Actual values: (NIL T T T T T T T)` — so they are not arm64
defects. That leaves **two** genuine port defects in the whole suite, both in
float printing.

Cleared since the last number, all build-confirmed: `.SPprogvsave`'s Floyd walk
was missing the loop-top nil test, so every even-length `progv` values list
signalled `XIMPROPERLIST` (5 tests — and the odd/even split predicts the passes
too); the kernel's `misc_ref`/`misc_set` legs ignored the complex-double-float
pad word that `arm64-arch.lisp:259-261` declares and every Lisp-side writer
honours (2); `%ALLOCATE-LIST` was wholly undefined, reachable only via
`MAKE-LIST` at size >= 65536 (1); `misc-ref-c-bit-fixnum` had no template (3);
and `EXPT.ERROR.4-7` needed an FPSR poll at the integer-power checkpoint, since
`(expt <float> <integer>)` is inline multiplication and never reaches `#_pow`
(4).

cyclistmass:
**Correction to my earlier comment: `COMPILE-FILE.2` and `.2A` are OUR bugs, not upstream's.**

I claimed they "fail identically on x86-64 CCL 1.12.2, so they are not arm64
defects." That was wrong, and I would rather flag it than leave it standing.

What I did now: ran the **full** ANSI suite under stock `lx86cl64` 1.12.2 on a
tree reset with `git clean -fdx` before each run. Twice, identically:

```
:TOTAL 21679 :FAILED 18
LOAD-TRUENAME.1  LOAD-PATHNAME.1  WARN.4
COMPILE-FILE.5 .6 .6A .7 .8 .9 .10 .11 .12 .13 .14 .16 .18 .19 .20
```

**`COMPILE-FILE.2` and `.2A` are not in that list.** They pass on x86 and fail
on arm64, so they are port defects. Both check that `COMPILE-FILE` returns the
right `warnings-p`/`failure-p` when the file produces style-warnings (`.2`) or
full warnings (`.2A`) — compiler diagnostics, nothing to do with pathnames.

The bad claim came from a *targeted* A/B that ran those two deftests outside
the suite, where they lose the aux setup `compile-file-test` depends on and can
fail for that reason alone. The in-suite result is the authority; I propagated
the targeted one without re-testing it in place.

**Corrected status: 21677 / 21679, with two genuine arm64 defects remaining**,
not zero. Also worth noting the earlier "19 failures" I would have reported was
residue — a stale `scratch/foo.txt` and untracked `COMPILE-FILE-TEST-LP.LSP`,
`LDTEST.LSP`, `TEMP.DAT` from previous runs. Only `ENSURE-DIRECTORIES-EXIST.8`
was caused by it; the other 18 are real on 1.12.2.

Those 18 all pass on our arm64 build — but ours is 1.13-derived from the `arm64`
branch, so the likely explanation is that they were fixed between 1.12.2 and
1.13 rather than anything to do with the port. I mention it only so the numbers
in this thread are comparable, not as a claim.

cyclistmass:
**Withdrawing the comment above: `COMPILE-FILE.2`/`.2A` are not arm64 defects, and the suite now passes with zero failures.**

I corrected myself once here already and got it wrong the second time too. The reason is worth more than the number, so:

**`RT::*TEST-VERBOSE*` decides those two tests on any conforming lisp.** `ansi-tests/rt.lsp:45` is `(defvar *test-verbose* t)`, so the suite's default is `T`. Then:

- `compile-file.lsp:35` muffles compiler warnings **only** when `*test-verbose*` is NIL —
  `(when (and (typep c 'warning) (not *test-verbose*)) (muffle-warning c))`
- `compile-file.lsp:50`'s `v1` reduces, for tests that pass neither `:print` nor `:verbose`, to **`(string= str "")`** — the stream captured around `compile-file` must be *empty*.

So with `:verbose t` the warning is not muffled, CCL prints it into that captured stream, and `v1` is NIL. Only `COMPILE-FILE.2` (`:expect-style-warnings t`) and `.2A` (`:expect-warnings t`) deliberately provoke warnings, so only those two flip. My arm64 run was at `:verbose t`; the x86 run I compared it against was at `:verbose nil`. The difference I published as "two arm64 defects" was the difference between two measurement cells.

Measured on both platforms, pristine tree (`git clean -fdxq` per run), `:compile t`, `DO-TESTS` at REPL toplevel:

| ANSI failures | RT `:verbose` | arm64 (this series) | stock x86-64 CCL 1.12.2 |
|---|---|---|---|
| | **NIL** | **0 / 21679** | **1** — `COMPILE-FILE.16` |
| | T | 2 — `CF.2`, `.2A` | 3 — `CF.16`, `CF.2`, `.2A` |

Ten runs per side in the `NIL` cell, no variance in count or failing-name set.

Also relevant to the `18` I quoted earlier: that run was inside an enclosing `LOAD`, and there `CF.2`/`.2A` **pass** even at `:verbose t`. Both axes gate the same quantity — whether warning text reaches the captured stream. It gets there only if the warning is neither muffled (`:verbose t` required) nor deferred to an enclosing compilation unit (toplevel required). That also looks like the explanation for `COMPILE-FILE.{5,6,6A,7,8,11,12,13,14,18}`, which fail inside a load and pass at toplevel — though I have not instrumented `with-compilation-unit` to confirm it.

Two things I am **not** claiming. `COMPILE-FILE.16` failing on 1.12.2 and passing here is more likely a 1.12.2→1.13 difference than anything about arm64. And a green ANSI suite is not a finished port — it rarely declares element types at `speed 3 / safety 0`, so it never reaches the open-coded array path, where there is a real bug: `arm642-vref1`'s `is-64-bit` `CASE` has no default arm, so a compiled `(signed-byte 64)` `AREF` emits nothing and returns the *index*. `arm642-vset1` has the matching gap on the write side (its `is-64-bit` branch emits `misc-set-{c-,}double-float` unconditionally). Both are in `arm642.lisp` as it stands, not introduced by this series; `ppc2.lisp:1382-1432` and `:2020-2045` have the shape, and I will send that separately.

cyclistmass:
Superseded by #564.

That PR is the same work as one tree based at `9fb4783` instead of split
Lisp/runtime at `9c61574`. The split here is what made this one unbuildable on
its own — ten kernel files were reaching our build through a build script rather
than through the series, so the series alone couldn't produce a kernel. All ten
are now patches, and #564 is gated on it: the sandbox is the pinned tree plus
the series and nothing else. It builds, boots, and runs ANSI 21679/21679 and
`ccl.lsp` 243/243.

Leaving this open rather than closing it, in case the history here is useful to
you — happy to close it if you'd rather.

cyclistmass:
Closing this in favour of #564, which carries the same work as one buildable tree based at 9fb4783.

PR #561, opened 2026-07-28
arm64: the Linux/arm64 kernel runtime and arm64-only Lisp sources (the other half of #560)

cyclistmass:
The other half of the Linux/arm64 port: the kernel runtime and the arm64-only
Lisp sources. #560 carries the Lisp-side *changes* to files you already have;
this carries the files you don't. **35 files, +19,983/−358, and 26 of them are
new** — so the two PRs are disjoint and can land in either order.

Based at `9c61574`, same as #560.

This is also the direct answer to "do I have access to the files needed to build
your Linux/arm64 ccl?" — with #560 and this, you do. Nothing is held back.

### The ff-call / callback design you described

We built it, and independently arrived at your invariant: **every non-scannable
cstack region is covered by a u64-vector whose element count, through
`skip_over_ivector`, lands exactly on the next boundary.** In `spentry-E-ffi.s`:

```
count  = ((OLD - sp) + node_size) >> node_shift
header = (count << num_subtag_bits) | subtag_u64_vector
```

pushed as the pair `{header, OLD}` at `sp-16`. OLD and sp are both 16-aligned by
AAPCS64, so `count*8 + 8 == OLD - (sp-16)` exactly and one
`lsl #(num_subtag_bits - node_shift)` does both shifts. Nested transitions
restore the *enclosing* boundary on the way out — without that, a
lisp→ffcall→callback→lisp→ffcall→return sequence leaves the boundary naming
popped stack and the next walk starts in dead memory.

**Where we differ, and it's the interesting part.** You publish the boundary *on
the stack*, shrinking the vector's count by 4 to expose a lisp frame whose
`savefn`/`savelr` are zeros until the frame is real. We keep the boundary in the
**tcr** — `tcr.last_lisp_frame`, which the GC reads in `normalize_tcr`'s ff-call
branch. Same guarantee, but the GC never *discovers* the boundary by walking, so
there is no ordering window: no interval where a half-built frame is visible and
no dependency on the store order of `savefn`/`savelr`. Yours keeps all the state
on the stack, which is tidier for backtrace. If you want one shape across both
ports I would rather adopt yours than maintain two — but the ordering constraint
in your third paragraph is the part I would want covered by a test that actually
GCs inside that window, because it is invisible when it is wrong.

Agreed on loading FPR args in Lisp and saving only x0-x7 for `.SPffcall`.

Not covering the C region cost us a real bug, which is the strongest argument
for the invariant: the marker walk climbed three real lisp frames, ran into the
C region above them, guessed through an x29 hop and four zero words, then read a
spilled `0.9d0` (`0x3feccccccccccccd`) as an ivector header and strode 1.8e16
words.

### The commits

1. **kernel runtime** — `spentry-{A..E}.s` (the subprimitive bodies), asmutils,
   the linuxarm64 platform header, `arm64_print.c`, and the GC/exceptions work.
   Also populates the subprimitive jump table: **130 of 132 rows**, where the tip
   ships 119 of 123 as `.quad 0`. Without it every compiler-emitted
   `ldr xN,[rcontext,#sptab+8*i]; blr xN` jumps to address 0.
2. **`level-0/ARM64/`** — 11 new LAP sources.
3. **`level-1/` + `lib/`** — callback support, error-signal, trap-support,
   threads-utils, backtrace.
4. **`linuxarm64`: make the kernel build.** The Makefile has never been run and
   is stale ARM32 inheritance: the `.s.o` rule assumes m4 where the arm64 asm
   layer is cpp (`imports.s` is the genuine m4 exception and keeps its own
   rule), `-mfloat-abi=` is not a valid aarch64 option, `-T ./armlinux.x` names
   a nonexistent linker script, `SPOBJ` needs the five new objects, `COBJ` wants
   `arm64_print.o` rather than the ARM32-tag-scheme `arm_print.o`, and `SPINC`
   still lists `arm64-constants.s`, which you removed at `f067047`. Plus
   `bits.h`, `albt.c` (`sp` is not in `regs[]` on arm64) and the `imports.s` PTR
   ladder.
5. **`thread_manager.c`** — `tcr->last_lisp_frame` was initialized under
   `#ifdef ARM`, so on arm64 it stayed 0 for the life of the thread. That is the
   boundary above. ARM32 needs it for the same reason (marker frames + linear
   walk); PPC does not, having a backlink chain.
6. **`arm64-constants.h`** — `gc.h:103` keys `is_symbol_fulltag` on
   `#ifdef fulltag_symbol`. x8664 defines it as a real macro; here it is a
   `DEFCONST`, i.e. an enum member, invisible to `#ifdef` — so arm64 silently
   compiled the `fulltag_misc` branch and the predicate matched nothing, ever.
   On a full GC, GCTWA then neither rescued worthy package-itab symbols nor
   scrubbed dead ones, and `compact_dynamic_heap` forwarded itab references to
   unmarked symbols. Heap corruption after the first full GC following heavy
   interning — in practice an in-image `compile-file`. gdb-confirmed:
   `compact_dynamic_heap -> node_forwarding_address`, `tag_n 7`.

### Two things that are your call

- **`arm64-spentry.s` renames your `misc_ref` spentry** to
  `misc_ref_upstream_sketch`. Yours is self-described as "just for playing with
  the uuos" and our `spentry-B` carries a complete body, so they collide at
  link time. Renaming yours is presumptuous — say which you want and we'll do it
  the other way round.
- **`spentry-E-ffi.s:572` keeps an `#error`** for the unported Darwin syscall
  convention (`svc #0x80` + carry-flag error protocol). It is `#ifdef DARWIN`,
  so it never fires on Linux.

### Verification, stated honestly

This branch has **not** been build-verified standalone. Our boxes build the
equivalent tree, but they get there by a script that applies these same changes
as `sed`s to a checkout, so "our build is green" is evidence about the content,
not about this branch. I am verifying the branch itself next and will report
here either way. Where the port gets to with this content plus #560: the ANSI
suite runs to completion at **21679 executed, 20 failures, no exclusions**.

Two build gotchas that will otherwise cost you an afternoon: the host CCL used
for the cross-compile must know the `aapcs64-ff-call` nx1 operator (a pristine
1.12.2 does not — a 1.13 from your `arm64` branch does), and a working install
is six parts, not two — omit `arm64-headers64/` and it dies with a bare SIGSEGV
because `CDB-OPEN` fails inside the error path.

### DIFF
--- a/doc/porting/linuxarm64-build.md
+++ b/doc/porting/linuxarm64-build.md
@@ -1,0 +1,98 @@
+# Building CCL for Linux/arm64
+
+Notes from getting the linuxarm64 target to build and boot. Assumes this
+branch (or the two PRs that produced it) is applied; with those in place the
+build is the ordinary CCL build, and what follows is only the parts that are
+not obvious.
+
+## You need two machines (or one plus a cross environment)
+
+- **an x86-64 Linux host** to build the heap image, and
+- **a Linux/arm64 machine** to build the kernel and run the result.
+
+The image cannot be built natively on arm64: building it runs the
+cross-compiler, which needs a working host CCL. The kernel is the only piece
+built natively on the target.
+
+## The host CCL must be new enough
+
+The host CCL that runs the cross-compile has to know the `aapcs64-ff-call`
+nx1 operator. **A pristine 1.12.2 does not**, and the cross-compile dies
+partway through with an obscure failure. A 1.13 built from the `arm64` branch
+(`da35a7d` or later) works.
+
+## Kernel
+
+```
+cd lisp-kernel/linuxarm64
+make
+```
+
+The Makefile in this branch has been corrected for aarch64; before that it
+was ARM32 inheritance and had never been run. In particular the `.s.o` rule
+builds with `$(CC) -x assembler-with-cpp` rather than m4, because the arm64
+assembly layer is cpp-based — `imports.s` is the one genuine m4 file and
+keeps a rule of its own.
+
+## An install is six parts, not two
+
+This is the step that wastes an afternoon. A runnable install needs:
+
+```
+armcl64                 the kernel
+arm64-boot-*.image      the heap image
+level-1.la64fsl         the startup fasl
+bin/                    level-1 fasls
+l1-fasls/               more of the same
+arm64-headers64/        the FFI interface databases
+```
+
+Copy only the kernel and the image and it dies with a **bare SIGSEGV that
+looks like a port bug and is not**: startup opens `level-1.la64fsl` relative
+to the install, `CDB-OPEN` then needs `arm64-headers64/`, and the failure
+happens *inside the error path* —
+
+```
+Interface file .../arm64-headers64/libc/records.cdb does not exist
+  -> Unknown foreign type: :<D>L_INFO
+  -> FATAL (cold load): unbound variable *PRINT-PPRINT-DISPATCH*
+```
+
+`strace` names it in one run; a register dump does not.
+
+Run with the install directory as cwd — the image resolves `ccl:` and its
+startup fasl relative to itself:
+
+```
+cd /your/install && ./armcl64 --image-name ./arm64-boot-16m16.image
+echo '(print (list :ok (+ 1 2) (sin 1.0d0)))' | ./armcl64 --image-name ./arm64-boot-16m16.image
+```
+
+Expect `(:OK 3 0.8414709848078965D0)`.
+
+## Floating point on this target
+
+AArch64 defines the FPCR trap-enable bits (`IOE`/`DZE`/`OFE`/`UFE`/`IXE`),
+but implementing trapped floating-point exceptions is **optional**, and they
+are RAZ/WI on at least Neoverse N1 — writing `0x1f00` to FPCR reads back
+`0x0`. So there is no SIGFPE to field the way `ppc-exceptions.c` does, and
+enabled-exception detection polls the cumulative FPSR flags at checkpoints
+instead.
+
+Two consequences worth knowing before debugging anything float-related:
+
+- **FPSR exception flags are cumulative.** Any code that captures them for a
+  foreign call must zero FPSR *before* the call, not only after it —
+  otherwise the captured word carries every flag raised since the previous
+  foreign call, including unrelated inline Lisp arithmetic, and the wrong
+  callee gets blamed. That produced spurious `FLOATING-POINT-OVERFLOW` from
+  `log`, `exp`, `sin` and `atan`.
+- Inline floating-point arithmetic therefore cannot signal on such a part.
+  `(expt <float> <integer>)` is repeated multiplication and never reaches
+  `pow`, so it needs its own checkpoint if it is to conform.
+
+## Status
+
+With this branch applied, the Dietz ANSI suite runs to completion at **21677
+of 21679 passing, no exclusions** — the two remaining failures reproduce
+identically on x86-64 CCL 1.12.2.
--- a/level-0/ARM64/arm64-array.lisp
+++ b/level-0/ARM64/arm64-array.lisp
@@ -1,0 +1,577 @@
+;;; -*- Mode: Lisp; Package: CCL -*-
+;;;
+;;; arm64-array.lisp — Wave-8 DRAFT port of vendor/ccl/level-0/PPC/ppc-array.lisp
+;;; PPC64 LINE-PORT (source: vendor/ccl/level-0/PPC/ppc-array.lisp)
+;;; Target: Matt Emerson upstream arm64 (low-tag) design, pin d71a5ad.
+;;; Per-line citations: "; ppc:NNN" = line NNN of vendor/ccl/level-0/PPC/ppc-array.lisp.
+;;;
+;;; CLASS-DISPATCH RESTRUCTURE (W8-D70): Matt's design has NO 8-bit ivector
+;;; class.  His classes (arm64-arch.lisp:89-91) are 64-bit (#b1101),
+;;; 32-bit (#b1100) and OTHER (#b0101); s8/u8/s16/u16/bit/complex-double-
+;;; float-vector all live in the OTHER class.  PPC64's four-way class
+;;; dispatch (64/32/8/residue) is re-keyed here: fulltag → 64/32, then the
+;;; OTHER residue dispatches on the header SUBTAG.  Per-subtag arms are
+;;; line-ported unchanged.
+;;;
+;;; NODEHEADER TEST (W8-D71): PPC64 tests the 2-bit lowtag against
+;;; lowtag-nodeheader.  Matt has no lowtags, but both nodeheader fulltags
+;;; (#b0110/#b1110, arm64-arch.lisp:62,70) share the 3-bit lisptag
+;;; tag-nodeheader (#b110, arm64-arch.lisp:52), so the test becomes
+;;; (and x tagmask) == tag-nodeheader.
+;;;
+;;; STATUS: DRAFT — not assembled; ledger in wave8-array-clos-report.md.
+
+(in-package "CCL")
+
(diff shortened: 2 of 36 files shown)


cyclistmass:
Measured update, since both PRs quote an older number.

With this content applied, the ANSI suite now runs to completion at
**21675 of 21679 passing — 99.98%, four failures, zero exclusions**
(`:EXCLUDED NIL`, `:EXCLUSIONS-NOT-FOUND NIL`, `EXIT:0`). It was 20 failures
this morning.

The four:

```
PRINT.SHORT-FLOAT.RANDOM    PRINT.SINGLE-FLOAT.RANDOM
COMPILE-FILE.2              COMPILE-FILE.2A
```

**`COMPILE-FILE.2` and `.2A` fail identically on x86-64 CCL 1.12.2**, with a
byte-identical `Actual values: (NIL T T T T T T T)` — so they are not arm64
defects. That leaves **two** genuine port defects in the whole suite, both in
float printing.

Cleared since the last number, all build-confirmed: `.SPprogvsave`'s Floyd walk
was missing the loop-top nil test, so every even-length `progv` values list
signalled `XIMPROPERLIST` (5 tests — and the odd/even split predicts the passes
too); the kernel's `misc_ref`/`misc_set` legs ignored the complex-double-float
pad word that `arm64-arch.lisp:259-261` declares and every Lisp-side writer
honours (2); `%ALLOCATE-LIST` was wholly undefined, reachable only via
`MAKE-LIST` at size >= 65536 (1); `misc-ref-c-bit-fixnum` had no template (3);
and `EXPT.ERROR.4-7` needed an FPSR poll at the integer-power checkpoint, since
`(expt <float> <integer>)` is inline multiplication and never reaches `#_pow`
(4).

cyclistmass:
Correction to my mail: I referred you to a `BUILD.md`, which lives in our own tree and is no use to you. The build notes are now in this branch as **`doc/porting/linuxarm64-build.md`**, rewritten for someone who has these commits and none of our tooling — with them applied it is just the ordinary CCL build, so the file only covers the parts that are not obvious: the image has to be cross-built on x86-64, the host CCL must know the `aapcs64-ff-call` nx1 operator (a pristine 1.12.2 dies partway through), an install is six parts rather than two, and the floating-point situation on parts whose FPCR trap enables are RAZ/WI.

cyclistmass:
**Correction to my earlier comment: `COMPILE-FILE.2` and `.2A` are OUR bugs, not upstream's.**

I claimed they "fail identically on x86-64 CCL 1.12.2, so they are not arm64
defects." That was wrong, and I would rather flag it than leave it standing.

What I did now: ran the **full** ANSI suite under stock `lx86cl64` 1.12.2 on a
tree reset with `git clean -fdx` before each run. Twice, identically:

```
:TOTAL 21679 :FAILED 18
LOAD-TRUENAME.1  LOAD-PATHNAME.1  WARN.4
COMPILE-FILE.5 .6 .6A .7 .8 .9 .10 .11 .12 .13 .14 .16 .18 .19 .20
```

**`COMPILE-FILE.2` and `.2A` are not in that list.** They pass on x86 and fail
on arm64, so they are port defects. Both check that `COMPILE-FILE` returns the
right `warnings-p`/`failure-p` when the file produces style-warnings (`.2`) or
full warnings (`.2A`) — compiler diagnostics, nothing to do with pathnames.

The bad claim came from a *targeted* A/B that ran those two deftests outside
the suite, where they lose the aux setup `compile-file-test` depends on and can
fail for that reason alone. The in-suite result is the authority; I propagated
the targeted one without re-testing it in place.

**Corrected status: 21677 / 21679, with two genuine arm64 defects remaining**,
not zero. Also worth noting the earlier "19 failures" I would have reported was
residue — a stale `scratch/foo.txt` and untracked `COMPILE-FILE-TEST-LP.LSP`,
`LDTEST.LSP`, `TEMP.DAT` from previous runs. Only `ENSURE-DIRECTORIES-EXIST.8`
was caused by it; the other 18 are real on 1.12.2.

Those 18 all pass on our arm64 build — but ours is 1.13-derived from the `arm64`
branch, so the likely explanation is that they were fixed between 1.12.2 and
1.13 rather than anything to do with the port. I mention it only so the numbers
in this thread are comparable, not as a claim.

cyclistmass:
**Withdrawing the comment above: `COMPILE-FILE.2`/`.2A` are not arm64 defects, and the suite now passes with zero failures.**

I corrected myself once here already and got it wrong the second time too. The reason is worth more than the number, so:

**`RT::*TEST-VERBOSE*` decides those two tests on any conforming lisp.** `ansi-tests/rt.lsp:45` is `(defvar *test-verbose* t)`, so the suite's default is `T`. Then:

- `compile-file.lsp:35` muffles compiler warnings **only** when `*test-verbose*` is NIL —
  `(when (and (typep c 'warning) (not *test-verbose*)) (muffle-warning c))`
- `compile-file.lsp:50`'s `v1` reduces, for tests that pass neither `:print` nor `:verbose`, to **`(string= str "")`** — the stream captured around `compile-file` must be *empty*.

So with `:verbose t` the warning is not muffled, CCL prints it into that captured stream, and `v1` is NIL. Only `COMPILE-FILE.2` (`:expect-style-warnings t`) and `.2A` (`:expect-warnings t`) deliberately provoke warnings, so only those two flip. My arm64 run was at `:verbose t`; the x86 run I compared it against was at `:verbose nil`. The difference I published as "two arm64 defects" was the difference between two measurement cells.

Measured on both platforms, pristine tree (`git clean -fdxq` per run), `:compile t`, `DO-TESTS` at REPL toplevel:

| ANSI failures | RT `:verbose` | arm64 (this series) | stock x86-64 CCL 1.12.2 |
|---|---|---|---|
| | **NIL** | **0 / 21679** | **1** — `COMPILE-FILE.16` |
| | T | 2 — `CF.2`, `.2A` | 3 — `CF.16`, `CF.2`, `.2A` |

Ten runs per side in the `NIL` cell, no variance in count or failing-name set.

Also relevant to the `18` I quoted earlier: that run was inside an enclosing `LOAD`, and there `CF.2`/`.2A` **pass** even at `:verbose t`. Both axes gate the same quantity — whether warning text reaches the captured stream. It gets there only if the warning is neither muffled (`:verbose t` required) nor deferred to an enclosing compilation unit (toplevel required). That also looks like the explanation for `COMPILE-FILE.{5,6,6A,7,8,11,12,13,14,18}`, which fail inside a load and pass at toplevel — though I have not instrumented `with-compilation-unit` to confirm it.

Two things I am **not** claiming. `COMPILE-FILE.16` failing on 1.12.2 and passing here is more likely a 1.12.2→1.13 difference than anything about arm64. And a green ANSI suite is not a finished port — it rarely declares element types at `speed 3 / safety 0`, so it never reaches the open-coded array path, where there is a real bug: `arm642-vref1`'s `is-64-bit` `CASE` has no default arm, so a compiled `(signed-byte 64)` `AREF` emits nothing and returns the *index*. `arm642-vset1` has the matching gap on the write side (its `is-64-bit` branch emits `misc-set-{c-,}double-float` unconditionally). Both are in `arm642.lisp` as it stands, not introduced by this series; `ppc2.lisp:1382-1432` and `:2020-2045` have the shape, and I will send that separately.

cyclistmass:
Superseded by #564.

That PR is the same work as one tree based at `9fb4783` instead of split
Lisp/runtime at `9c61574`. The split here is what made this one unbuildable on
its own — ten kernel files were reaching our build through a build script rather
than through the series, so the series alone couldn't produce a kernel. All ten
are now patches, and #564 is gated on it: the sandbox is the pinned tree plus
the series and nothing else. It builds, boots, and runs ANSI 21679/21679 and
`ccl.lsp` 243/243.

Leaving this open rather than closing it, in case the history here is useful to
you — happy to close it if you'd rather.

cyclistmass:
Closing this in favour of #564, which carries the same work as one buildable tree based at 9fb4783.

PR #565, opened 2026-08-03
xdump: size the cloned package tables from the target, not the host

cyclistmass:
Thanks for merging #564.

One patch, and it is **not an arm64 change** — `xdump/xfasload.lisp` is the shared
xloader, so every cross-built target carries this. It applies to `master`
unchanged (`git apply --check` clean there); I based it on `arm64` only because
that is where the merge left things. Happy to retarget.

`xload-clone-packages` sizes each cloned package's symbol tables from the
**cross-compile host's** corresponding package — `(length (car (uvref p 0)))` is
the length of the host's `pkg.itab` vector, and the `cddr` carries the host's
limit across too. But the clone is a *fresh* package that `xload-intern` then
populates from the cold-load fasls, so it inherits a size derived from a host
that has compiled all of CCL. Two consequences, both measured:

**The tables are enormously oversized.** On the linuxarm64 boot image, 32738
dumped slots holding 2423 symbols — 34.96% of dumped dynamic space is table, and
the empty part of it is 32.37%. Because the clone inherits the host's *limit*
along with its length, `%resize-htab` is never reached during the xload, so the
host's size lands in the image verbatim.

**Worse, it makes the boot image a function of the host session rather than of
the sources.** `target-xcompile-level-0` runs in the same process as `xfasload`,
so whether level-0 needed recompiling changes how many symbols the host has
interned. `touch`ing the level-0 sources of an otherwise untouched tree — same
revision, same `SOURCE_DATE_EPOCH` — moved one `%resize-htab` step in the host's
`KEYWORD` package and moved the dumped image by exactly `1308*8` bytes, while the
dumped symbol set and count stayed identical at 2423. That is also the
long-standing "first build after a tree reset is 12288 bytes larger" effect.

The fix sizes the clones the way `make-package` sizes a new package and lets
`%htab-add-symbol` grow them.

**Why it is safe rather than merely smaller:** the dumped sizes are not
load-bearing past the first instant of cold load. The image's initial
`%toplevel-function%` in `level-0/nfasload.lisp` runs `%resize-htab` over every
package's itab and etab *unconditionally*, before anything else interns, and
`%resize-htab` sizes from actual occupancy — so it shrinks as readily as it
grows, and it was already converting the old host-sized tables into very nearly
what this patch dumps directly (CCL's itab at 1870 symbols asks for 2339, which
`$primsizes` rounds to 2609, exactly what the patched cloner produces). No lookup
path becomes newly live: the target has always rehashed every dumped table with
its own `hash-pname` on every boot.

Effect on the linuxarm64 image: dumped slots 32738 → 3690, image 1576976 →
1347600 bytes, **−14.5%**. A cold build now reproduces a warm one bit-for-bit.

Boot-confirmed and bar-clean on a Neoverse-N1: image
`adb37b7fb8f9460f7f97c2498b2ac085`, ANSI 21679/0 with no exclusions and `ccl.lsp`
243/0, against a same-kernel same-fasl control built without the patch that gives
identical results. Since a table-modulus change could in principle make a present
symbol report as absent without crashing, symbol lookup was exercised rather than
inferred: 34462 accessible and 3857 external symbols round-tripped from table
iteration back through `find-symbol` with 0 misses, absent names correctly absent,
and forced rehashes of `KEYWORD` and `CCL` lost none of 1959 and 16654 entries.

The two initial sizes are `defparameter`s rather than literals so a port that
wants headroom in its boot image can raise them without editing the cloner.

### DIFF
--- a/xdump/xfasload.lisp
+++ b/xdump/xfasload.lisp
@@ -857,20 +857,89 @@
     (declare (fixnum n))
     (values (xload-save-string str n) str n new-p)))
 
+;;; The cloned packages that go into the image are FRESH packages: they start
+;;; out empty and are populated by XLOAD-INTERN as the cold-load fasls are read.
+;;; They used to be created with their symbol tables sized from the CROSS-COMPILE
+;;; HOST's corresponding package -- (length (car (uvref p 0))), i.e. the length of
+;;; the host's PKG.ITAB vector -- which has two consequences, both measured on a
+;;; linuxarm64 boot image.
+;;;
+;;; 1. The tables are enormously oversized, because the host has compiled all of
+;;;    CCL while the image holds only what level-0 interns.  Dumped slot counts
+;;;    vs symbols actually in them:
+;;;
+;;;      CCL      itab 24989 slots / 1870 symbols     etab  1153 /  78
+;;;      CL       itab    41 /    0                   etab  1153 / 345
+;;;      KEYWORD  itab    41 /    0                   etab  2609 / 105
+;;;      TARGET   itab  2609 /   25                   etab    41 /   0
+;;;      OS       itab    61 /    0                   etab    41 /   0
+;;;
+;;;    32738 slots for 2423 symbols.  At 8 bytes a slot that is 261904 of the
+;;;    749232 bytes of dumped dynamic space: 34.96% of that space is table, and
+;;;    the EMPTY part of it is (32738 - 2423) * 8 = 242520 bytes, 32.37%.  The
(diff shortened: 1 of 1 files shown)


xrme:
Please make this PR against master.

We don't really care much about optimizing the bootstrapping image.  But your argument about not needlessly changing the boot image due to irrelevant state in the host CCL is reasonable.

Plus, the proposed code changes are simple (and arguably an improvement in clarity).

cyclistmass:
Redone against `master` as #567 — thanks. Closing this one.

Issue #32, opened 2017-03-30
Vector streams external-format defaulting


davidmullen:
When using make-vector-output-stream, the default external-format is :DEFAULT. For with-output-to-vector, it's NIL.

xrme:
Thank you for the bug report.

PR #571, opened 2026-08-06
arm64: tstack push must not use a writeback store with Rt == Rn

cyclistmass:
This is the startup crash you hit running the linuxarm64 binaries under Linux on
Apple Silicon.

    Unhandled exception 4 at 0x300000034c9c, insn 0xf81d0f18 (neither udf nor brk)
    #<Function REFRESH-EXTERNAL-ENTRYPOINTS> <- #<Function RESTORE-LISP-POINTERS>

`0xf81d0f18` is `str x24, [x24, #-48]!` — a store with writeback whose transfer
register and base register are the same register. The ARM ARM makes that
CONSTRAINED UNPREDICTABLE (`Unpredictable_WBOVERLAPST`, C3.2.1) and UNDEFINED is
one of the permitted behaviours, so whether it executes at all is an
implementation choice. Your hardware takes the UNDEFINED option.

### How it was rooted

Not from the register dump. From the exception number: the kernel prints the
faulting word only for SIGILL or SIGTRAP (`arm64-exceptions.c:1989`), and a store
into mapped writable memory cannot raise SIGILL. So the fault had to be the
*encoding*, not the address.

GNU `as` says the same thing without any special hardware:

    wb.s:4: Warning: unpredictable transfer with writeback -- `str x24,[x24,#-48]!'

and assembles it to exactly `f81d0f18`.

### Where it came from

Three tstack-push vinsns — `make-stack-cons`, `macptr->stack`,
`make-stack-vcell` — transliterated from PPC64's

    (define-ppc64-vinsn make-stack-cons ... (stdu tsp -32 tsp))

where `stdu rX,-N(rX)` is the ordinary, well-defined stack push. AArch64 has no
equivalent single instruction: the store must name a register other than the base
it updates. The kernel `.s` files are clean; this is only in the compiler.

### The fix

`mov link, tsp` then `str link, [tsp, #-N]!`, so Rt != Rn. Same frame layout, and
the same backlink at `[tsp, #0]` that `discard-temp-frame` reads. Assembles with
no warning.

### On the GC-safety question

You raised that your two candidate fixes could leave a half-built frame visible
to an async interrupt. **This form does not have that problem.** The store and
the base update are still a single instruction, so an interrupt lands either
strictly before or strictly after the push — the GC never observes a partially
constructed tstack frame. No `pc_luser_xp` work is needed.

### Red-then-green, measured without your hardware

The check is whether the encoding is present in the built image at all:

- before: image `79a1606c` contains `f81d0f18` **× 33**
- after:  image `6488e51a` contains it **× 0**

Full builds are green on the fix: x86 cross-build `rc0=0 rc1=0 rc2=0`; on
Graviton, ANSI `21679/0` and `ccl.lsp` `243/0`, `EXIT:0`.

Worth noting for the record: this survived months of green ANSI runs on the
identical bytes, because the implementations we test on take the benign option.
A green suite says nothing about a CONSTRAINED UNPREDICTABLE encoding.

### DIFF
--- a/compiler/ARM64/arm64-vinsns.lisp
+++ b/compiler/ARM64/arm64-vinsns.lisp
@@ -3535,8 +3535,10 @@
 ;;; verbatim -- his kernel's tstack walker doesn't exist yet and must
 ;;; match this layout when it lands.
 (define-arm64-vinsn make-stack-cons (((dest :lisp))
-                                     ((car :lisp) (cdr :lisp)))
-  (str tsp (:@! tsp (:$ -32)))
+                                     ((car :lisp) (cdr :lisp))
+                                     ((link :u64)))
+  (mov link tsp)
+  (str link (:@! tsp (:$ -32)))
   (str xzr (:@ tsp (:$ 8)))
   (str xzr (:@ tsp (:$ 16)))
   (str xzr (:@ tsp (:$ 24)))
@@ -4137,9 +4139,10 @@
 ;;; macptr-header = his define-header (arch:709), fits movz.
 (define-arm64-vinsn macptr->stack (((dest :lisp))
                                    ((address :u64))
-                                   ((header :u64)))
+                                   ((header :u64) (link :u64)))
   (movz header (:$ arm64::macptr-header))
-  (str tsp (:@! tsp (:$ -48)))
+  (mov link tsp)
+  (str link (:@! tsp (:$ -48)))
   (str xzr (:@ tsp (:$ 8)))
   (str header (:@ tsp (:$ (+ 16 arm64::fulltag-misc arm64::macptr.header))))
   (str address (:@ tsp (:$ (+ 16 arm64::fulltag-misc arm64::macptr.address))))
@@ -4363,7 +4366,8 @@
 ;;;     (() ()) (ld tsp 0 tsp))
 ;;; -- pop the temp-stack frame by loading the BACKLINK, which every
 ;;; tstack frame stores at [tsp, #0] (our V3b make-stack-cons writes it
-;;; with (str tsp (:@! tsp (:$ -32))); same PPC64 stdu protocol --
+;;; with (mov link tsp) + (str link (:@! tsp (:$ -32))); same PPC64 stdu
+;;; protocol, but NOT PPC64's single stdu -- see make-stack-cons --
 ;;; RATIFY item "tstack frame shape" already queued with Matt).
 ;;; tsp = his x24 (arm64-asm.lisp:215); frames are 16-aligned so the
 ;;; offset-0 scaled LDR encodes.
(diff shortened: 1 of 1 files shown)


cyclistmass:
You were right about the GC window, and I was wrong when I said this form did not
have one. Second commit pushed (fast-forward, the first commit did not move).

Two things I got wrong. The store and the base update being a single instruction
makes the *push* atomic; it says nothing about the *frame*, which is what you
were actually talking about. And the window is worse than one instruction: the
vinsns set `tsp_frame.type = 0` — "contains nodes" — immediately after the push,
before any data slot is written, so a GC there scans stale temp-stack contents as
tagged nodes. All four walkers do `if (current[1] == 0) mark_simple_area_range(...)`.

PPC64 never had it, and shows the intended shape — `make-tsp-cons`,
`ppc64-vinsns.lisp:2106`:

```lisp
(stdu ppc::tsp -32 ppc::tsp)      ; push
(std  ppc::tsp 8 ppc::tsp)        ; type = tsp, non-zero => raw, GC skips
(stfd ppc::fp-zero 16 ppc::tsp)   ; fill data
(stfd ppc::fp-zero 24 ppc::tsp)
(std  ppc::rzero 8 ppc::tsp)      ; type = 0 => now scannable
```

We dropped the mark-raw store when transliterating. Which explains something you
may want to know independently of this PR: **that store is
`MARK_TSP_FRAME_INSTRUCTION`** — `0xF9000718`, `str x24,[x24,#8]`, storing the
non-zero tsp. Our vinsns emitted `0xF900071F`, `str xzr,[x24,#8]`, storing 0.
Different Rt, opposite meaning. So `pc_luser_xp`'s tsp clause has never fired on
a single frame it was written for.

Your `stp` suggestion is what I used, and it needs only one scratch — the old tsp
is the value being stored and is never 0:

```lisp
(mov link tsp)
(stp link link (:@! tsp (:$ -32)))   ; backlink + non-zero type, atomic
...fill data...
(str xzr (:@ tsp (:$ 8)))            ; type = 0 last
```

So you were right that no `pc_luser_xp` support is needed — though for the
opposite reason to the one I gave you: not because the old form was safe, but
because `stp` leaves no window to cover.

`macptr->stack` also stops clearing the type word. A macptr frame holds a foreign
address, not nodes, and PPC64's `macptr->stack` leaves it raw for the frame's
whole life; ours was marking it as containing nodes.

Verified on the artifact, since the suite cannot see this: the boot image's 33
`-48` pushes and 9 `-32` pushes became exactly 33 and 9 `stp` pairs, 0 single-str
pushes left, and the 9 remaining `str xzr,[x24,#8]` are the cons/vcell frames
where clearing the type word is correct — the 33 macptr frames no longer have it.
x86 cross `rc0=0 rc1=0 rc2=0`; on Graviton ANSI `21679/0`, `ccl.lsp` `243/0`.

GNU `as` takes `stp x12,x12,[x24,#-32]!` with no warning. Worth noting the twin
does warn — `ldp x12,x12,[x24],#32` gives "unpredictable load of register pair" —
so the pair-overlap rule is asymmetric between load and store.

PR #572, opened 2026-08-06
arm64: AAPCS64 record returns, HFA/HVA classification, callbacks, stack arguments, and natural operators

cyclistmass:
Eight commits completing the AAPCS64 side of the linuxarm64 FFI: record returns,
HFA/HVA classification, the callback direction, foreign-call **stack arguments**,
and the unboxed-`u64` operators the FFI paths need. Based on `arm64` at 933c9276.

Independent of #571 — the two branches merge cleanly in either order. I have
verified this branch also merges cleanly into `arm64` as it stands today (i.e.
with #571 in), so there is nothing to rebase.

### The commits

1. **interpreted `%FF-CALL`** — so a toplevel `EXTERNAL-CALL` works at all. Until
   this, `%ff-call` only existed as a compiled path.
2. **HFA/HVA classification + register-returned records** — the AAPCS64 rules for
   which aggregates come back in `v0`-`v3` / `x0`-`x1` rather than through memory.
3. **`>16`-byte non-HFA record returns in `x8`** — the AAPCS64 indirect result
   pointer, plus the `.SPffcall-indirect-result` entry.
4. **callback HFA arguments and register-returned records** — the lib-only half.
5. **callback kernel half** — reload `d0`-`d3`, capture the incoming `x8`, and
   move the callback index off `x8` so it survives.
6. **natural (unboxed `u64`) arithmetic and shift operators** — closes four
   operator gaps the paths above hit.
7. **stack arguments in foreign calls** — one NSAA area, both register classes.
8. **whole-composite stack arguments (C.3 / C.11)** and NSAA in the interpreted
   `%ff-call`.

Commits 7 and 8 are new since you last saw this branch; they are appended, so
1-6 are the same commits with the same shas. The rest of this description is
about them.

### 7. Stack arguments: the compiler was already right; `.SPffcall` was not

More than 8 GPR arguments (or more than 8 FP arguments) was not slow or subtly
wrong, it was **refused**:

```
CCL::COMPILER-BUG: aapcs64-ff-call: more than 8 GPR args (12)
                   -- stack-arg frame layout not ratified
```

That refusal is honest, because the codegen alone could not make such a call
work. But it is only half the story, and the half that was missing is small.

**The counting pass already did the AAPCS64 assignment.** Pass 1 already counts
overflow arguments into `nother-words`, `nother-words` is already included in
`single-words`/`total-words` so the c-frame is already sized for them, and pass 2
already stores arguments 9..N at param words 8, 9, 10… That block is already
contiguous, ascending, left-to-right, 8-byte slots, and 16-byte aligned. It is
already a valid NSAA.

**The one defect was that the callee never saw it.** `.SPffcall` kept SP at the
frame head across the `blr`, so a callee looking for its stack arguments at
`[SP]` read the c-frame's `u64`-vector header.

The fix is to step SP over `header + savedsp + params[0..7]` — `c_frame.size +
8*node_size`, the constant `.SPsyscall` in this same kernel already steps by —
immediately before the `blr`, so the overflow block sits **at** `[SP]` per
AAPCS64 5.4.2.

Cross-port, in the order I look:

* **PPC64 has no answer to copy, because it has no such problem.** In PowerOpen
  the stack parameters live in the *caller's* frame at a positive offset from
  the caller's SP, so `poweropen_ffcall` never has to move SP onto them.
* **x86-64 is the exact analog** and does exactly this. `_SPffcall` in
  `x86-spentry64.s` skips the header and frame-pointer words, pops the six GPR
  words, and calls with `rsp == &param[6]`; it survives the callee clobbering
  everything below the new `rsp` by keeping the one thing it still needs — the
  previous foreign SP — in callee-saved `%rbp`.

Everything the return path used to re-read from the frame head is therefore
hoisted into callee-saved registers before the call, because that whole region
is below the callee's incoming SP and dies there: `save1` holds the boundary
`lisp_frame`, `save0` the enclosing foreign boundary (whose park in param word 0
disappears entirely). `tcr.last_lisp_frame` now points at the published boundary
`lisp_frame` itself, which `mark_cstack_area` already accepts as a walk start
since it classifies a frame that begins with the marker.

Applied to all three variants: `.SPffcall`, `.SPffcall_return_registers`, and
`.SPffcall_indirect_result` (commit 3's).

### GC atomicity, again — three windows in the same seams

While auditing the new sequences instruction boundary by instruction boundary I
found three pre-existing windows of exactly the class you caught in #571 (one
instruction makes the *push* atomic, not the *frame*):

1. **The boundary `lisp_frame` was built with the real `fn`/`lr` stored while the
   frame was still covered by the `u64`-vector header** — where the GC can
   neither see nor forward them. A GC that moves the caller in that window leaves
   a stale `savefn`/`savelr` which the return path then reads back. The
   `ALLOC-C-FRAME` design comment already prescribes the right order (zeros while
   covered, publish via the count shrink, only then store the real `fn`/`lr`); the
   assembly never implemented it. Now it does, at all four sites — the three
   `ffcall` variants and `syscall`.
2. **The return path popped the NVRs and reloaded `allocptr`/`allocbase` while
   still at foreign valence**, where a GC forwards the vstack slot and the TCR
   slot but never the register. The tail now follows the order PPC64 and x86-64
   agree on: make every node register GC-valid, flip valence to lisp, *then* pop
   the (possibly forwarded) slots; `allocptr`/`allocbase` cross the flip as
   `VOID_ALLOCPTR` and are reloaded from the TCR after.
3. All four boxed NVRs are now spilled to the vstack across the call, before the
   boundary frame's `savevsp` is written, so the GC keeps and relocates them.

None of these three is caused by the stack-argument work; the SP move would have
widened (1) and (2), so they are fixed rather than inherited.

`.SPsyscall` gets fix 1 only. Its own tail still pops `save3`/`fn` before the
valence flip — the same window as (2) — which I have deliberately left alone
here: it is pre-existing, it is not widened by anything in this branch, and
`svc` does not clobber the user stack, so its SP/park scheme is safe as it
stands. Worth a separate look, and I am happy to send that separately.

One comment fix rides along: `arm642.lisp`'s header block documented the frame
as `[GP save 8][FP save 8][stack args M]`, which is not the layout — the
single/double-float staging area is *above* the stack arguments and is
variable-size. The comment now describes what the code does.

### 8. AAPCS64 never splits a composite across registers and the stack

Removing the blanket refusals exposed two shapes that the refusals had been
keeping honest, and both would otherwise have compiled to something silently
wrong:

* **An HFA whose members do not all fit in `v0`-`v7`** (C.2 fails): C.3 sends
  the **whole** aggregate to the stack, with natural member packing, and sets
  `NSRN := 8` so every later scalar FP argument stacks too. Member-wise
  `:single-float` decomposition cannot express this: a stacked float member is 4
  bytes, while C.5 widens a scalar single to an 8-byte slot.
* **A 2-doubleword record with exactly one GPR left** (C.10 fails): C.11 wastes
  the odd register — no argument may straddle the register file and the stack —
  and sets `NGRN := 8`, then C.12/C.13 copy the whole record to the NSAA.

Both are expressed in `expand-ff-call` with its running per-class counts: dummy
register-class argforms burn the registers the ABI wastes (dead content per the
callee's prototype, and they keep the backend's own counters in lockstep so
later scalars flow through the overflow arms from commit 7 untouched), and the
aggregate payload is emitted as `ceil(size/8)` words of a new backend argspec
`:stack-doubleword` — one raw 8-byte NSAA slot, never a register. That is C.13's
memory copy, bit for bit. A tail of 4 bytes or fewer is read narrow so the copy
never reads past the object. Composites with 16-byte alignment (C.8/C.12) are
refused precisely rather than silently mis-slotted.

**The burn is the AAPCS64-specific part**, and it is where I deviate from the
x86-64 donor rather than follow it: SysV's MEMORY class burns no registers, so
`x8664-backend.lisp`'s otherwise identical running-counter shape has nothing
equivalent. PPC64 has no analog at all, since PowerOpen reserves a memory slot
per parameter. Noted in the comments.

The interpreted `%ff-call` marshals NSAA arguments too now — one shared stack
offset past the 8 GPR words, advanced in source order, frame sized
`(+ 8 nstack)`. The stock x86-64 `%ff-call` in `x86-def.lisp` is the line donor,
re-counted for AAPCS64.

#### A latent defect this made live

The interpreted `%ff-call` used to pass a constant frame size, so
`alloc-variable-c-frame` had **never been emitted in this port's history**.
Sizing the frame `(+ 8 nstack)` reached it, and it did not work:

* its wired `prevsp` temp was declared `(:imm imm1)`, a mode
  `allocate-temporary-vreg` has no entry for, so the cross-build died at xload
  with `:IMM is not (MEMBER :CRF :U8 … :LISP …)`;
* with that corrected, the vinsn's **unwired** `size` temp was allocated `x1` —
  the same register its **wired** `prevsp` temp claims — so `mov prevsp,sp`
  clobbered the just-computed size and `sub sp,sp,size` zeroed SP. Observed
  directly in a boot core: `sp == 0` at the faulting `stp`.

Fixed by wiring all three temps `:u64` to `imm0`/`imm1`/`imm2`, which leaves the
`imm0`/`imm1` `stp` encoding `pc_luser_xp` pattern-matches unchanged. The
general hazard is worth knowing about independently of this branch: the temp
allocator does not exclude a wired temp's register from the unwired pool, so any
template mixing wired and unwired temps can collide.

Separately, and pre-existing: the *compiled* `%ff-call` special form accepts
integer word-count argument specs through `nx1` that `arm642` has no case for,
and dies with `Unknown gpr mode name: NIL`. `expand-ff-call` never emits them,
so it is reachable only from hand-written `%ff-call` forms. Not touched here.

### Why this mattered in practice

An aarch64 Oracle client driven through CCL's FFI needs functions taking 12, 11
and 10 integer arguments. They were not slow or subtly wrong — they were
un-callable, and one of them defines an output variable, so no query could run
at all. That is what prompted commit 7.

### Testing

A dedicated test with a C fixture, 21 cases, each argument carrying a distinct
bit so a wrong slot names itself: the 12/11/10-integer shapes; 9 and 10 doubles;
an `int,double,int` interleave that kills any per-class marshaller; GPR overflow
under live FP registers; a stacked single-float (C.5 widening); a pure-FP
single/double NSAA mix; 12-integer calls under forced GC; the C.3 and C.11
aggregate cases; the interpreted-`%ff-call` cases; and two negative controls
(an HFA that fits, a 16-byte record that fits).

* **Watched red first, by message and not merely by failure.** Before commit 7:
  10/10 fail, every one in `COMPILE` with the exact refusal text. Before commit
  8: 9 of 21 fail, the five aggregate cases with the expander's exact refusal
  texts and the four interpreted cases with `%ff-call`'s, while the other 12 —
  including both negative controls — pass in the same run.
* **Checked against stock x86-64 CCL as the oracle**: the same test file and
  fixture pass `21/0` there, so the test is testing the ABI and not this port.
* **Green after**: `21/0`, on both the bootstrap and the purified image.
* **Cross-build** `rc0=0 rc1=0 rc2=0`; **ANSI `21679/0`** and **`ccl.lsp`
  `243/0`** on Graviton with the new pair, `EXIT:0`. Both the image and the
  kernel md5 moved, so the change demonstrably reached the binaries.

The caveat I flagged for #571 still applies and applies more here: ANSI does not
reach FFI, callbacks, or the GC-atomicity paths by construction, so `21679/0` is
not coverage evidence for this branch. The 21-case test and the red controls are.

### One non-arm64 file, deliberately

`compiler/nx1.lisp` gets a nine-line change in commit 3. The existing check
special-cases `:registers` — the x8664/darwinppc64 result-register capture buffer
— and rejects duplicates. AAPCS64 needs a second spelling, `:indirect-result`,
for the same conceptual thing (the record result), so the check becomes a
`memq` over both and the error message names whichever keyword was duplicated.
No existing behaviour changes: `:registers` alone still behaves identically.

If you would rather that hunk arrived separately, say so and I will split it out.
Same offer for commits 7 and 8 if you would rather review the kernel SP change
on its own.

### DIFF
--- a/compiler/ARM64/arm64-arch.lisp
+++ b/compiler/ARM64/arm64-arch.lisp
@@ -860,6 +860,11 @@
 (defconstant callback-frame.fp-save-offset -64)
 (defconstant callback-frame.savelr-offset -152)
 (defconstant callback-frame.stack-args-offset 64)
+;;; Incoming x8 (the AAPCS64 6.9 indirect result-area pointer, live when
+;;; the callback returns a >16-byte non-HFA record), captured by
+;;; _SPcallback into the padding word of its foreign-sp stash pair
+;;; (spentry-E-ffi.s `stp imm0, save2').
+(defconstant callback-frame.x8-save-offset -248)
 
 (defmacro define-header (name element-count subtag)
   `(defconstant ,name (logior (ash ,element-count num-subtag-bits) ,subtag)))
--- a/compiler/ARM64/arm64-arch.lisp
+++ b/compiler/ARM64/arm64-arch.lisp
@@ -1406,7 +1406,12 @@
                           ;; catch/throw/unwind-protect/progv cluster (w10;
                           ;; spentry-C-bind-catch-throw.s:319/326/334/1663/
                           ;; 343/440/550, spentry-B:478/1044)
-                          (defsubprim .SPnmkunwind)))))))
+                          (defsubprim .SPnmkunwind)
+                          ;; AAPCS64 6.9 indirect-result (x8) ff-call
+                          ;; variant, for >16-byte non-HFA record-by-value
+                          ;; returns (16m71 HFA lane); body in
+                          ;; spentry-E-ffi.s, sptab index 132
+                          (defsubprim .SPffcall-indirect-result)))))))
 
 ;;; The extension above rebinds the arm64::*subprims* VARIABLE, but
 ;;; ccl::subprim-name->offset resolves through the arch STRUCT's
--- a/compiler/ARM64/arm64-vinsns.lisp
+++ b/compiler/ARM64/arm64-vinsns.lisp
@@ -300,8 +300,32 @@
 (define-arm64-vinsn (alloc-variable-c-frame) (()
                                               ((n-c-args :lisp))
                                               ((header (:u64 #.arm64::imm0))
-                                               (size :u64)
-                                               (prevsp (:imm #.arm64::imm1))))
+                                               ;; ALL THREE temps are WIRED
+                                               ;; (:u64 mode, not the :imm
+                                               ;; CLASS -- a wired temp is
+                                               ;; allocated by MODE and
+                                               ;; allocate-temporary-vreg has
+                                               ;; no :imm entry; the original
+                                               ;; (:imm imm1) type-errored the
+                                               ;; FIRST time this vinsn was
+                                               ;; ever emitted, since the old
+                                               ;; %ff-call passed a constant
+                                               ;; frame size and only
+                                               ;; ALLOC-C-FRAME was live).
+                                               ;; size is wired too because
+                                               ;; the temp allocator handed
+                                               ;; an UNWIRED size the same x1
+                                               ;; the wired prevsp claims --
+                                               ;; mov prevsp,sp then clobbered
+                                               ;; size and `sub sp,sp,size'
+                                               ;; zeroed SP (observed in a
+                                               ;; boot core, sp=0 at the stp).
+                                               ;; imm0/imm1 must keep their
+                                               ;; wiring: pc_luser_xp
+                                               ;; recognizes the stp below by
+                                               ;; its exact encoding.
+                                               (size (:u64 #.arm64::imm2))
+                                               (prevsp (:u64 #.arm64::imm1))))
   (add size n-c-args (:$ '6))        ;+ header + prevsp + 4-word frame
   (add size size (:$ (:apply 1- arm64::dnode-size))) ;round byte size up...
   (and size size (:$ (:apply - arm64::dnode-size)))  ; ...to a dnode boundary
--- a/compiler/ARM64/arm64-vinsns.lisp
+++ b/compiler/ARM64/arm64-vinsns.lisp
@@ -6489,8 +6489,9 @@
 
 ;;; ============ set-c-arg / set-single-c-arg / set-double-c-arg ============
 ;;; argnum is a WORD index into the param area (param0 = word 0); the
-;;; handler assigns words 0-7 to GPR args, 8.. to overflow, then FP
-;;; staging.  Byte offsets are 16+8k: 8-aligned, scaled-STR encodable
+;;; handler assigns words 0-7 to GPR args, 8.. to the NSAA stack-arg
+;;; block (overflow args of BOTH register classes, in source order --
+;;; AAPCS64 has ONE next-stacked-argument address), then FP staging.  Byte offsets are 16+8k: 8-aligned, scaled-STR encodable
 ;;; through the whole u16const range.  Little-endian: singles store/
 ;;; load at offset+0 (PPC's +4 was big-endian) -- v2 donor deviation
 ;;; kept.  Register width (X/S/D form) follows the operand class, as
--- a/compiler/ARM64/arm64-vinsns.lisp
+++ b/compiler/ARM64/arm64-vinsns.lisp
@@ -6541,6 +6541,22 @@
   (ldr temp (:@ rcontext temp))
   (blr temp))
 
+;;; ============ ff-call-return-registers ============
+;;; ff-call variant for record-by-value returns: same frame and entry
+;;; contract as ff-call, plus a regbuf macptr in arg_y into which the
+;;; subprim stores every AAPCS64 result register on return
+;;; ({x0-x7 @ 0..56, d0-d7 @ 64..120}; `spentry ffcall_return_registers',
+;;; spentry-E-ffi.s).  x86-64 pairs the same way:
+;;; (define-x8664-subprim-call-vinsn (ff-call-return-registers)
+;;; .SPffcall-return-registers), x8664-vinsns.lisp; PPC64's donor is
+;;; .SPpoweropen-ffcall-return-registers (ppc64-vinsns.lisp).
+(define-arm64-vinsn (ff-call-return-registers :call :subprim) (()
+                                                               ()
+                                                               ((temp (:u64 #.arm64::imm1))))
+  (movz temp (:$ (:apply arm64::subprimitive-offset ".SPffcall-return-registers")))
+  (ldr temp (:@ rcontext temp))
+  (blr temp))
+
 ;;; ============ eep.address ============
 ;;; PPC64 ppc64-vinsns.lisp:3829: load slot 1 (the address) of an
 ;;; external-entry-point gvector, trap if NIL (unresolved eep -- PPC's
--- a/compiler/ARM64/arm64-vinsns.lisp
+++ b/compiler/ARM64/arm64-vinsns.lisp
@@ -6890,6 +6890,49 @@
                                       (y :u64)))
   (eor dest x y))
 
+;;; ============ natural (unboxed u64) arithmetic and shifts ============
+;;; Demand: acode-rewrite.lisp:241-246 rewrites + and - to %natural+ and
+;;; %natural- whenever BOTH operand types are subtypes of target-natural-type,
+;;; with no reader conditional -- so ordinary portable source that declares
+;;; (unsigned-byte 64) operands at speed 3 produces these operators on arm64,
+;;; and without a handler the compile fails.  natural-shift-left/right arrive
+;;; the same way for constant shift counts.  ANSI cannot reach any of it (a
+;;; declared-type open-coded path), which is why 21679/0 never saw it.
+;;;
+;;; PPC64 ppc64-vinsns.lisp:3834-3849 (%natural+ add / %natural- sub) and
+;;; :2960-2966 (natural-shift-left/right).  PPC64 expresses the shifts with
+;;; rldicr because it has no dedicated 64-bit shift-immediate;
+;;; ;;; ARM64-DEVIATION: AArch64 has lsl/lsr with an immediate (UBFM aliases),
+;;; so the rotate-and-mask encoding has no analogue and is not imitated.  The
+;;; count is :u8const on both ports, matching PPC64's assumption that nx1 has
+;;; already established a constant amount in range.
+;;;
+;;; The -c (constant-operand) variants PPC64 carries are deliberately NOT
+;;; ported: the arm64 natural family established by %natural-logand/-logior/
+;;; -logxor above has no -c forms either, and their handlers always materialise
+;;; both operands into imm registers.  Matching the neighbouring arm64
+;;; convention keeps one spelling of the family and avoids an imm12-range
+;;; question that buys only one movz.
+(define-arm64-vinsn %natural+ (((dest :u64))
+                               ((x :u64)
+                                (y :u64)))
+  (add dest x y))
+
+(define-arm64-vinsn %natural- (((dest :u64))
+                               ((x :u64)
+                                (y :u64)))
+  (sub dest x y))
+
+(define-arm64-vinsn natural-shift-left (((dest :u64))
+                                        ((src :u64)
+                                         (count :u8const)))
+  (lsl dest src (:$ count)))
+
+(define-arm64-vinsn natural-shift-right (((dest :u64))
+                                         ((src :u64)
+                                          (count :u8const)))
+  (lsr dest src (:$ count)))
+
 ;;; ============ ivector-typecode-p / gvector-typecode-p ============
 ;;; Demand (16m28, pin advance to 33e61e6): his arm642-ivector-typecode-p
 ;;; and arm642-gvector-typecode-p (arm642.lisp:6188/6195) emit them, and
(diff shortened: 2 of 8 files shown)


xrme (changes requested):
There seem to be some run-time problems with this code.

I created a test branch test-572 at the current tip of arm64
(ba84eb2e).  I then checked out the PR branch, and merged it into my
test-572 branch.

After doing (rebuild-ccl :clean t), I noticed that doing (quit)
segfaulted.  Additionally, when trying to (rebuild-ccl :clean t)
from the freshly-rebuilt lisp (with the changes from #572 included),
the lisp segfaulted after about half a dozen files.

Appended is an edited log of what I did.


$ git checkout arm64
$ git checkout -b test-572
$ gh pr checkout 572
branch 'arm64-ffi-hfa-records' set up to track 'cyclistmass/arm64-ffi-hfa-records'.
Switched to a new branch 'arm64-ffi-hfa-records'
$ git checkout test-572
$ git merge arm64-ffi-hfa-records
Auto-merging compiler/ARM64/arm64-arch.lisp
Auto-merging compiler/ARM64/arm64-vinsns.lisp
Merge made by the 'ort' strategy.
 compiler/ARM64/arm64-arch.lisp      |  12 +-
 compiler/ARM64/arm64-vinsns.lisp    | 107 +++++-
 compiler/ARM64/arm642.lisp          | 241 +++++++++++--
 compiler/nx1.lisp                   |   9 +-
 level-0/ARM64/arm64-def.lisp        | 349 +++++++++++++++++++
 level-1/arm64-callback-support.lisp |  21 +-
 lib/ffi-linuxarm64.lisp             | 657 +++++++++++++++++++++++++++++-------
 lisp-kernel/arm64-spentry.s         | 199 ++++++++---
 lisp-kernel/spentry-E-ffi.s         | 272 +++++++++++----
 9 files changed, 1591 insertions(+), 276 deletions(-)


$ ./arm64cl -n
Clozure Common Lisp Version 1.13 (v1.13-322-gba84eb2e) LinuxARM6464

For more information about CCL, please see https://ccl.clozure.com.

CCL is free software.
It is distributed under the terms of the Apache License, Version 2.0.
? (rebuild-ccl :clean t)
Rebuilding Clozure Common Lisp using Version 1.13 (v1.13-322-gba84eb2e) LinuxARM6464
Path to source code: #P"/Users/rme/leda/ccl/"
;Compiling "/Users/rme/leda/ccl/lib/systems.lisp"...
;Loading #P"/Users/rme/leda/ccl/bin/systems.la64fsl"...
[...]
? (quit)
$ ./arm64cl
Clozure Common Lisp Version 1.13 (v1.13-331-g12d4cc75) LinuxARM6464

For more information about CCL, please see https://ccl.clozure.com.

CCL is free software.
It is distributed under the terms of the Apache License, Version 2.0.
? (quit)
Segmentation fault

After trying that a couple of times with identical results, I tried
rebuilding:

$ ./arm64cl
Clozure Common Lisp Version 1.13 (v1.13-331-g12d4cc75) LinuxARM6464

For more information about CCL, please see https://ccl.clozure.com.

CCL is free software.
It is distributed under the terms of the Apache License, Version 2.0.
? (rebuild-ccl :clean t)
Rebuilding Clozure Common Lisp using Version 1.13 (v1.13-331-g12d4cc75) LinuxARM6464
Path to source code: #P"/Users/rme/leda/ccl/"
;Compiling "/Users/rme/leda/ccl/lib/systems.lisp"...
;Loading #P"/Users/rme/leda/ccl/bin/systems.la64fsl"...
;Compiling "/Users/rme/leda/ccl/lib/compile-ccl.lisp"...
;Loading #P"/Users/rme/leda/ccl/bin/compile-ccl.la64fsl"...
;Compiling "/Users/rme/leda/ccl/compiler/nxenv.lisp"...
;Loading #P"/Users/rme/leda/ccl/bin/nxenv.la64fsl"...
;Compiling "/Users/rme/leda/ccl/compiler/nx.lisp"...
; Including "/Users/rme/leda/ccl/compiler/nx-basic.lisp"...
; Including "/Users/rme/leda/ccl/compiler/lambda-list.lisp"...
; Including Segmentation fault

cyclistmass:
Thanks for the detailed reproduction — the log is exactly what we needed.

We have not reproduced this yet. Our testing is on Graviton4 under Linux arm64, where the branch runs 21679/0. That green result cannot rule out what you hit.

One question first. What hardware is this running on? I ask because of the defect you found in June: our tstack-push vinsns emitted `str x24,[x24,#-48]!`, a store whose transfer register is its base register. ARM ARM C3.2.1 makes that CONSTRAINED UNPREDICTABLE and permits UNDEFINED. Apple silicon takes the UNDEFINED option and Graviton takes a benign one, so we stayed green on the identical bytes for months. If your machine is Apple hardware, this failure has the same shape: it appears for you and not for us, on the same commits.

Two things we will do now:

1. Run `(rebuild-ccl :clean t)` twice in sequence with this branch applied, then `(quit)` from the rebuilt lisp. Our self-rebuild evidence comes from a pristine clone before these changes went in, so we have not tested that path with them.
2. Scan the branch for writeback encodings where Rt equals Rn.

If you have a core file from the `(quit)` segfault, the faulting instruction and the frame would settle question 1 quickly. We will report back either way.

Mauro

xrme:
> > One question first. What hardware is this running on? I ask because of the defect you found in June: our tstack-push vinsns emitted `str x24,[x24,#-48]!`, a store whose transfer register is its base register. ARM ARM C3.2.1 makes that CONSTRAINED UNPREDICTABLE and permits UNDEFINED. Apple silicon takes the UNDEFINED option and Graviton takes a benign one, so we stayed green on the identical bytes for months. If your machine is Apple hardware, this failure has the same shape: it appears for you and not for us, on the same commits.

My machine is a Linux VM on a Mac Studio, so Apple Silicon.

> Two things we will do now:
> 
> 1. Run `(rebuild-ccl :clean t)` twice in sequence with this branch applied, then `(quit)` from the rebuilt lisp. Our self-rebuild evidence comes from a pristine clone before these changes went in, so we have not tested that path with them.
> 2. Scan the branch for writeback encodings where Rt equals Rn.
> 
> If you have a core file from the `(quit)` segfault, the faulting instruction and the frame would settle question 1 quickly. We will report back either way.

If your're not seeing it, I'll at least attach the debugger to see what I can see, and maybe generate a core file.

cyclistmass:
We ran both things we said we would, and one of the results points away from
this branch.

**The writeback scan.** We scanned the kernel built from your current arm64 tip
with this PR rebased on it (92 raw hits), and our own pinned-tree kernel (79).
Every hit maps to a non-executable ELF section — mostly DWARF, plus one each in
`.eh_frame`, `.debug_line` and `.strtab`; 77 of the 79 are non-ALLOC and never
mapped at run time. Executable sections: zero, by two independent decode passes.
We watched the check go red first: injecting `f81d0f18` into a copy of `.text`
reports one CODE hit. So there is no Rt==Rn writeback in executable code on
either build.

**The double rebuild.** With the branch applied, `(rebuild-ccl :clean t)` then
`(quit)`, then the same again from the rebuilt image: rc 0 both times,
REBUILD-COMPLETE, 159 files each pass. We did not reproduce your `(quit)`
segfault. That is consistent with Apple-only visibility, and your core file would
settle it.

**The part you will care about more.** We ran the same sequence on pristine
upstream with none of this PR applied, as a control. The second rebuild
segfaults there too: rc 139 after four files, during `compiler/nx0.lisp`. Two
independent runs, both cored. With our full patch queue applied it stops
happening.

So your "and then rebuild-ccl fails after about half a dozen files" looks like a
pre-existing defect in upstream rather than a regression from this branch.

Two limits on that claim. Our control ran at a tree 23 commits after the
`ba84eb2e` you based `test-572` on, so it is not your exact base. And we have not
identified which patch in our queue masks it — that is a 26-wide bisect we have
not run. The crash decodes to compiled Lisp at heap PC `0x30000017c378`, on
`ldr x11,[x10,x0]` feeding `str x11,[x25,#-8]!`, with x0 holding a garbage
offset.

Mauro DiBenedetto <maurodibenedetto@gmail.com>

xrme:
The rebuild-ccl crash after a few files was a different bug.  Two bugs, perhaps: both involving icache management.

In my 80e636c2, I flush the whole compacted region again. This undoes your earlier optimization, and might be a performance regression for you, but this caused reproducible failures on Apple hardware.  I also added a missing make-code-executable in arm64-lap-generate-code in my 8ef03ec.  (That one's a bad miss and is my fault.)

I'll re-test these proposed changes now.

xrme:
It looks like I may have been tripped up by bootstrapping order.  In order for this to work, it's necessary to rebuild the lisp before rebuilding the lisp kernel.  Rebuilding the lisp kernel first, and then using that lisp kernel with an old image will cause the segfault-on-quit.

xrme (approved):
After building in the correct order (image, lisp kernel), (quit) exits cleanly, and a second rebuild works.

PR #573, opened 2026-08-06
Portable: REBUILD-CCL toolchain fixes, plus arm64 arms for sockets, backtrace, machine-version, heap reports, leaks, describe and ELF

cyclistmass:
Five portable commits against `master`. Three add `#+arm64-target` arms to
per-target dispatches that currently have none; two fix `REBUILD-CCL` toolchain
handling and are not arm64-specific at all.

Nothing here changes existing behaviour on any current target — the arm64 arms
are reader-conditionalised, and the `compile-ccl` changes only affect which
program gets exec'd and when the failure is reported.

### The two `compile-ccl` fixes (not arm64)

**`"make clean"` must go through `MAKE-PROGRAM`.** `rebuild-ccl` calls
`(make-program)` everywhere except the `make clean` invocation, which is a
hardcoded `"make"`. On any platform where the GNU make binary is called
something else — the BSDs, where it is `gmake` — a `:clean t` rebuild runs the
wrong make.

**Check for make before `REBUILD-CCL` does the work.** The kernel build is the
*last* thing `rebuild-ccl` does, so `compile-ccl` and `xload-level-0` have
already run by the time a missing toolchain is discovered. `run-program` cannot
exec make, nothing reaches the output stream, and the user gets a bare
"generic OS error in fork/exec" after a long wait.

This is not hypothetical: someone with no C toolchain installed ran
`rebuild-ccl` on a fresh machine and got exit 255 with no usable diagnosis. The
check runs before the Lisp work and names what to install. Only `make` is
pre-checked, deliberately — every other tool is invoked *by* make and names
itself when absent ("m4: command not found"). The error text points at the
existing "Kernel Build Prerequisites" section of the manual, which already
covers the C compiler, `as`, `ld`, `m4` and GNU make.

### The three arm64 arms

**`sockets`** — `%verify-socket-buffer` checks the ivector subtag against
per-target bounds and has arms for ppc / x8632 / x8664 / arm. Without an arm64
arm the `and` yields NIL and every valid `(unsigned-byte 8)` buffer is rejected
by `report-bad-arg`.

**`backtrace`** — `MAKE-VSP-STACK-RANGE` and `MAKE-CSP-STACK-RANGE` had no arm64
definitions. Worth noting the structural point, because the obvious move is
wrong: arm64 sits in **PPC's** position here, not ARM32's. arm64 catch frames are
misc-tagged temp-stack uvectors with an explicit `csp` slot holding a
control-stack `lisp_frame` pointer, so both functions read that slot the way the
ppc arms do. ARM32's arms compute `CATCH-FRAME-SP` by pointer arithmetic instead,
because ARM32 has no temp stack and stack-conses its catch frames.

**`machine-version`** — the `/proc/cpuinfo` key to look for was unset on arm64,
so `machine-version` returned NIL. An aarch64 kernel emits neither x86's
"model name" nor ARM32's "Hardware": `arch/arm64/kernel/cpuinfo.c`'s `c_show()`
writes processor / BogoMIPS / Features / CPU implementer / CPU architecture /
CPU variant / CPU part / CPU revision, and "Hardware" comes from `arch/arm/`
only. "CPU part" is the MIDR_EL1 part number — the core model, and the most
specific identifier that is always present.

---

## Appended: five more arm64 arms in portable files

Same shape as the three above — a per-target dispatch with no arm64 arm, which
compiles clean and degrades silently. **The reachability of each differs a lot,
so it is stated per patch rather than as one claim.**

### `misc: populate heap type names on arm64` — exercised

`*heap-utilization-vector-type-names*` has x8664, ppc64 and 32-bit initializers
and no arm64 one, so all 256 entries stay NIL and a heap-utilization report on
arm64 names nothing. The arm follows the ppc64 form using arm64's own low-tag
geometry: select on the five header fulltags from `arm64-arch.lisp`, index with
`(ash i (- arm64::ntagbits))`. arm64's tables call their raw storage classes
`function-vector` and `symbol-vector`, so the display names are normalized back
to ppc64's `function` and `symbol`.

**Measured red-then-green on two arm64 images built one patch apart, same
kernel: `0` symbol names of 256 slots → `80`.** 80 is also exactly what stock
x86-64 CCL reports, which is the right cross-check — this arm is the x8664 arm's
counterpart and should name the same number of typecodes.

*Unrelated, but you should have it:* `heap-utilization` itself **core-dumps on
arm64** (SIGSEGV, not a signalable condition). Measured on images with and
without this patch, same kernel, so it is pre-existing and independent. The same
call returns normally on stock x86-64 1.12.2. Separate defect; I have not chased
it yet.

### `leaks: hide arm64 raw symbol vectors` — compiles and loads, filter not exercised

`transitive-referencers` removes implementation vectors so a caller sees the user
object rather than both it and its backing storage. arm64 has the separately
mapped `symbol-vector` class, as x8664 does, but was not in the filter. arm64 has
no `function-vector` mapping, so that reader conditional stays x8664-only.

`library/leaks.lisp` compiles clean for linuxarm64 and is loaded into the image
(`bin/leaks.la64fsl`), but nothing in ANSI or `ccl.lsp` runs a leak report, so
the filter itself has not been observed doing its job.

### `describe: give arm64 the function-inspector PRIN1-VALUE arm` — compiles clean, not exercised

The method is defined twice and **both** arms are suppressed on arm64
(`#+x86-target`, `#+ppc-target`), so arm64 falls through to the generic
`prin1-value`, whose body is `(prin1 value stream)`: every disassembly line of an
inspected function prints as a raw cons instead of its text.

The x86 arm is the donor rather than the ppc one, deliberately: it is data-shape
logic with no ISA content, while the ppc arm additionally filters `:comment`
entries — a statement about `ppc-disassemble`'s output shape, not about the ISA.
If arm64 disassembly turns out to emit `:comment` entries, the right move is to
widen this arm again with evidence, not to inherit a filter for entries nobody
has seen.

**Not exercised.** `ccl::prin1-value` is not fbound in a deployed image on arm64
*or* on stock x86-64 — the inspector is not loaded by default — so the check
skips on both targets and even definedness could not be asserted.

### The two `elf` patches — correct source fixes to a feature arm64 cannot currently build

⚠️ **State plainly what these are not.** `library/elf.lisp` **cannot be compiled
for linuxarm64 at all** today, so neither patch has been compiled, loaded or run
on this target. Please do not read them as tested.

The cause is independent of both patches, and I positive-controlled that rather
than asserting it. The file opens with `(use-interface-dir :elf)`, and our arm64
interface database has exactly **one** directory where x86-64 ships eight:

```
arm64-headers/     -> libc
x86-headers64/     -> elf gl gmp gnome2 gnustep gtk2 jni libc
```

so `#_elf_errmsg`, about 30 lines in, has nothing to resolve against. Compiling
the **unpatched file at the pin** under a control name fails **identically, at
the identical byte range** (`Read error between positions 1370 and 1523`); both
patched sites are after that point and are never reached. That is a gap in our
interface database, not something these patches cause or fix.

They are still worth having, because both defects are real in the source and one
of them is an arity loss:

**`collect-elf-static-functions`** has exactly two definitions, `#+x86-target`
and `#+(or arm-target ppc-target)`, so on arm64 it is simply undefined — and it
is the entry point the file uses to enumerate purified static functions before
writing them as ELF symbols, the mechanism that lets gdb and perf name a lisp
function in a saved image. The x86 arm is the donor because it uses the portable
`%map-areas` walker rather than hand-walking the readonly area with
`area.code`/`area.low` and `fixnumshift`, and its inner `#-x8664-target` branch
already hands arm64 the right type.

**`st_value`, `st_size` and `e_machine`.** One `setf` with five place/value
pairs loses *both* its arm64 values — all three `st_value` arms and all three
`st_size` arms vanish. Two vanishings leave an even token count, so `setf` does
not become a syntax error and nothing warns; it degenerates to
`(setf (pref p ...st_value) (pref p ...st_size))`. The buffer came from
`#_calloc`, so every arm64 ELF symbol would come out **address 0, size 0** —
a well-formed symbol table that is useless to any external symbolizer.

Two notes on that one:

* Not `(ash (uvsize cv) word-shift)`. An arm64 code-vector is
  `ivector-class-32-bit` but `word-shift` is 3, so that spelling **doubles** the
  size. **The ppc64 arm has exactly that bug today** — `library/elf.lisp:269` is
  right on ppc32 (`n*4`) and 2× wrong on ppc64 (`n*8`). Independent of arm64 and
  worth a look. This patch uses `subtag-bytes`, which derives the element width
  from the subtag.
* `new-elf-file-header` takes **four required parameters**, and its only call
  site passes `machine` from a five-arm dispatch with no arm64 arm. On arm64 all
  five vanish and the call is made with **three** arguments — so
  `write-elf-symbols-to-file`, the only public entry point of the whole feature,
  cannot run at all the moment the file becomes compilable. That is why the two
  are sent as a pair: the first alone leaves the feature broken.

One dependency I could not check and am not hiding: whether the generated
linuxarm64 interface database exports `EM_AARCH64`. The pin names it nowhere, and
the interface DB is generated binary rather than text. The value (183) is
certainly right; if the constant is absent the failure is **loud and at read
time**, and the remedy is regenerating the linux ELF interface directory — the
same gap as above.

### What the arm64 bar does and does not say for these five

Cross-build `rc0=0 rc1=0 rc2=0`; ANSI **21679/0** and `ccl.lsp` **243/0** on
Graviton, with the image md5 moving between the two builds so the patches
demonstrably reached the binary. For `misc` that sits alongside a direct
red-then-green measurement. For the other four it is a **no-regression** result
and nothing more — `leaks`, `describe` and `elf` are in the set ANSI cannot reach
by construction, which is why these defects survived every 21679/0 run.

---

### Correction, added after this PR was opened: the ELF collector commit above was inert on arm64

The commit "elf: define COLLECT-ELF-STATIC-FUNCTIONS on arm64", already in this branch, is wrong. I have appended a new commit that corrects it, rather than rewrite a commit you may already have read. I am sorry for the round trip.

That commit gave arm64 the `#+x86-target` arm. That arm scans the readonly area for objects of type `FUNCTION`. It is right for x86-64, where `purify` moves the function-vector. On arm64 `purify` moves the CODE VECTOR instead, and leaves the function object in dynamic space. No readonly object satisfies the predicate, so the collector returns NIL.

`write-elf-symbols-to-file` then writes a 584-byte file. Its `.symtab` holds only the mandatory index-0 NULL entry. `readelf` reports `Machine: AArch64` correctly, and then zero symbols. **The feature did nothing at all as sent. It was not merely imperfect.**

Measured on a linuxarm64 image, with both candidate bodies run verbatim in one image:

| measurement | value |
|---|---|
| the arm as sent returns | **0** functions |
| the corrected arm returns | **12674** functions, all `FUNCTION`, one per distinct code vector |
| objects in `:readonly` after `purify` | 78682 |
| of those, `(typep o 'function)` | **0** |
| of those, `subtag-code-vector` (typecode 65) | 12764 |
| of those, `subtag-function` (typecode 96) | **0**, absent from the census entirely |
| the same predicate over `:dynamic` | 15812 functions |

The last row is the check on the instrument. The predicate matches 15812 objects in the same image. So the zero tells us where the objects live. It does not tell us that the walk is broken.

The correction moves arm64 onto the `#+(or arm-target ppc-target)` arm. That arm needs one arm64 value: `(code-vector-index #+ppc-target 0 #+arm-target 1 #+arm64-target 0)`.

To repair the x86 arm in place is not an option. On arm64, `function-vector-to-function` expands to `%function-vector-to-function`. That helper is an identity with `(trap-unless-typecode= arg arm64::subtag-function)`. A code vector traps.

The other elf commit in this PR is "elf: arm64 st_value, st_size and e_machine". It already computes `st_value` from `(uvref f 0)`, so it already assumed that the code vector is the purified object. The two commits contradicted each other. That is the clearest sign that the first one was a mistake and not a judgment call.

**Verification boundary.** The arm64 counts above come from hardware. I did not measure the ELF symbol table end to end. That needs `library/elf.lisp` compiled with the correction, and we publish no image that carries that file. This change also makes a code path live on arm64 for the first time. I therefore expect a close reading of the resulting symbol table to find more.

---

## Appended: correct the sockets arm — the interval is not the class on arm64

The sockets commit above gave arm64 the interval spelling the 32-bit ports
use: accept every subtag from `min-8-bit-ivector-subtag` to
`max-8-bit-ivector-subtag`. That spelling is exact on ppc32, x8632 and arm32
because their 8-bit subtags are contiguous. It is not exact on arm64. arm64
interleaves the ivector classes through the low fulltag bits
(`compiler/ARM64/arm64-arch.lisp` on the arm64 branch), so the interval
[#xD4, #xE4] also contains `s32-vector` (#xD5) and `s64-vector` (#xDD).

Measured on hardware against live typecodes: `(make-array 4 :element-type
'(signed-byte 32))` has typecode #xD5, `(signed-byte 64)` has #xDD, and the
interval admits both. A vector that passes wrongly then reaches
`%copy-ivector-to-ptr`, where `offset` and `size` count bytes — but
`verify-socket-buffer` validated them against `LENGTH`, which counts
elements. The socket sends the wrong bytes, silently.

The correction compares the two 8-bit subtags exactly. That set is the type
that `report-bad-arg` names. The arm64 uses of the min/max bounds elsewhere
in the tree compare only after a dispatch on `ivector-class-other-bit`, so
the exact comparison is also the form consistent with how those constants
are used.

One observation, not changed here: x8664 has the same interleaved layout,
and its arm in this function has the same property — `s32-vector` (#xD9) and
`s64-vector` (#xDA) fall inside its interval [#xD7, #xE7] (master
`compiler/X86/X8664/x8664-arch.lisp`). Measured on stock CCL 1.12.2
LinuxX8664: `verify-socket-buffer` accepts a `(signed-byte 32)` vector
(typecode #xD9) and a `(signed-byte 64)` vector (#xDA), and rejects
`(unsigned-byte 32)` (#xE9). Happy to send a separate fix for that if you
want one.


xrme:
I cherry-picked these to master:
ce0ad05d  compile-ccl: check for make before REBUILD-CCL does the work
c3f018d1  compile-ccl: "make clean" must also go through MAKE-PROGRAM

I cherry-picked these to arm64:
3c1739c8  misc: MACHINE-VERSION returns NIL on arm64
fe923822  misc: populate heap type names on arm64 (measured 0→80)
71bd3d72  backtrace: MAKE-VSP/CSP-STACK-RANGE for arm64

I also went ahead and took these to arm64 as well:
5b0486be  leaks: hide arm64 raw symbol vectors
b39107ec  describe: function-inspector PRIN1-VALUE arm

The others I'm going to decline here.  These are all relevant to elf.lisp.  When elf.lisp is more or less ready, I'd like a separate PR for it, please.
6f5b50fa  collect-elf-static-functions (superseded; author says inert)
35627e81  st_value / st_size / e_machine
a8de1dc8  collector correction

PR #577, opened 2026-08-07
arm64: add the missing WITH-C-FRAME p2 handler

cyclistmass:
## arm64: add the missing `WITH-C-FRAME` p2 handler

> **Stacked on #572.** This branch is #572's *first* commit
> (`arm64: interpreted %FF-CALL, so toplevel EXTERNAL-CALL works`) plus one new
> commit. Only the second is new here. That dependency is real and not
> cosmetic: this patch's trailing context is the `WITH-VARIABLE-C-FRAME` handler
> that commit adds, and the patch does not apply at the bare branch point —
> I checked with `git apply --check` rather than assuming. Once #572 merges this
> collapses to a single commit.

`with-c-frame` and `with-variable-c-frame` are a pair, and every other back end
defines both:

| back end | fixed-size | variable-size |
|---|---|---|
| `compiler/PPC/ppc2.lisp` | `:8922` `ppc2-with-c-frame` | `:8930` `ppc2-with-variable-c-frame` |
| `compiler/ARM/arm2.lisp` | `:9592` `arm2-with-c-frame` | `:9600` `arm2-with-variable-c-frame` |
| `compiler/X86/x862.lisp` | `:11626` `x862-with-c-frame` | `:11632` `x862-with-variable-c-frame` |
| `compiler/ARM64/arm642.lisp` | **absent** | present |

arm64 had neither until the variable-size half was added for the interpreted
`%ff-call`; this adds its fixed-size peer.

The body is the x86-64 shape rather than the ppc or arm32 one, because those
`ecase` over an EABI c-frame variant that arm64, like x86-64, does not have.
`alloc-c-frame` with a constant size, `$undo-arm64-c-frame` and
`arm642-undo-body` are all already in use by the variable-size handler, so this
introduces nothing new.

### ⚠️ Reachability, stated plainly: this is NOT tested

**linuxarm64 cannot exercise it.** The only in-tree caller of `with-c-frame` is
the objc-bridge (`objc-bridge/objc-runtime.lisp:3207,3223`), which is
Darwin-only. There is no path from a linuxarm64 test to this handler, so please
do not read the green bar below as coverage.

What *was* observed, red-then-green, is that the handler **installs**: two arm64
images built one patch apart, same kernel, asked the p2 dispatch table by the
handler's own `function-name`.

| check | before | after |
|---|---|---|
| `arm642-with-c-frame` present in the dispatch table | **absent** | present |
| `arm642-with-variable-c-frame` (pre-existing sibling, control) | present | present |
| a deliberately non-existent handler (negative control) | absent | absent |

The sibling and the negative control are in the set on purpose: an earlier
version of that probe used `fboundp` and reported *both* handlers missing, which
is impossible — `defarm642` installs into a table via `svset` and never fbinds
the symbol. Two reds where one was impossible means the instrument is wrong, not
the patch.

Also: cross-build `rc0=0 rc1=0 rc2=0`, ANSI **21679/0**, `ccl.lsp` **243/0**, and
the image md5 moved between the two builds — so the change reached the binary and
regressed nothing. That is the whole of the runtime claim.

### Why it is worth landing anyway

`compiler/ARM64/arm642.lisp` is the back end for **both** arm64 targets. On
darwinarm64 there is no p2 handler for the operator at all, so there is no
locative for it and the objc-bridge cannot compile — a loud failure, but a
blocking one. Latent on linux, blocking on Darwin.

### DIFF
--- a/compiler/ARM64/arm642.lisp
+++ b/compiler/ARM64/arm642.lisp
@@ -9756,6 +9756,24 @@
       (! %current-frame-ptr target)))
   (^))
 
+;;; WITH-C-FRAME: allocate a c-frame of the default size around BODY.
+;;; The fixed-size half of the pair below; every other back end defines
+;;; both (ppc2.lisp ppc2-with-c-frame / arm2.lisp arm2-with-c-frame /
+;;; x862.lisp x862-with-c-frame), and this one is byte-for-byte the
+;;; x86-64 shape because arm64, like x86-64, has a single c-frame vinsn
+;;; and no EABI variant to ecase over.
+;;;
+;;; The only in-tree caller is the objc-bridge (objc-runtime.lisp:3207,
+;;; 3223), which is Darwin-only -- so this is latent on linuxarm64.  It
+;;; is NOT latent on darwinarm64: arm642.lisp is the back end for both
+;;; arm64 targets, so without this handler the bridge has no locative
+;;; for the operator and simply cannot compile there.
+(defarm642 arm642-with-c-frame with-c-frame (seg vreg xfer body &aux
+                                                 (old-stack (arm642-encode-stack)))
+  (! alloc-c-frame 0)
+  (arm642-open-undo $undo-arm64-c-frame)
+  (arm642-undo-body seg vreg xfer body old-stack))
+
 ;;; WITH-VARIABLE-C-FRAME: allocate a c-frame of SIZE param words around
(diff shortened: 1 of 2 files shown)


xrme:
I cherry-picked 00450c08ecfec997fc1d28db8ff503b42a796764 to arm64 as e32b662f9.

#572 was merged already, so I hope that I am right not to bother with dealing with the merge conflicts in the other commit.

PR #578, opened 2026-08-07
arm64: .SPsyscall's return path reloads lisp registers at foreign valence

cyclistmass:
## arm64: `.SPsyscall`'s return path reloads lisp registers at foreign valence

This is the separate patch offered at the end of #572:

> `.SPsyscall` gets fix 1 only. Its own tail still pops `save3`/`fn` before the
> valence flip — the same window as (2) — which I have deliberately left alone
> here: it is pre-existing, it is not widened by anything in this branch […]
> Worth a separate look, and I am happy to send that separately.

### The defect

`spentry syscall` comes back from the `svc` and, while `tcr.valence` still says
`TCR_STATE_FOREIGN`, reloads `allocptr`/`allocbase` from the TCR, reads `lr` back
out of the boundary `lisp_frame`, and pops `save3` and `fn` off the vstack. Only
after all five reloads does it store `TCR_STATE_LISP`.

A GC in that window is another thread's — this thread is foreign, so it is
suspended and walked, not stopped at a safe point. It forwards the vstack slots
and the TCR slots. It does **not** forward this thread's registers: at foreign
valence the collector walks from `tcr.last_lisp_frame`, and `mark_xp`/`forward_xp`
do not cover a foreign context's register file. So every value loaded into a
register *before* the flip keeps its pre-move address, and the flip publishes
those stale addresses as live lisp state. `fn` is the worst of them — a stale
function pointer, which is what the caller returns into.

This is the same shape fixed in the three ff-call bodies in #572 (`spentry
ffcall`, `ffcall_return_registers`, `ffcall_indirect_result`). `.SPsyscall` is
the fourth foreign-transition subprim and the only one still doing it.

### Cross-port: both donors already have the order this adopts

* `ppc-spentry.s` `poweropen_syscall` — `str(imm4,tcr.valence(rcontext))` comes
  **before** `vpop_saveregs`, before `ldr(loc_pc,lisp_frame.savelr(sp))` and
  before `ldr(fn,lisp_frame.savefn(sp))`.
* `x86-spentry64.s` `_spentry(syscall)` — `movq $TCR_STATE_LISP,rcontext(...)`
  comes **before** `pop %fn` and the rest of the pops.

arm64 was the outlier. The return path is rewritten to ppc64's order: retreat SP
onto the boundary `lisp_frame`, make every node register GC-valid (nil the
arg/temp set, nil `fn`, zero the raw-address NVRs, set `allocptr`/`allocbase` to
`VOID_ALLOCPTR`), flip to `TCR_STATE_LISP`, and only then hand back the enclosing
boundary, read `lr` from the frame at `[sp]`, pop the NVRs and reload the
allocation pointers.

### Two related fixes ride along, because they are the same defect

* **The prologue spilled only `fn` and `save3`.** `save0`, `save1` and `save2`
  hold lisp values too, and a value that crosses the foreign window only in an
  NVR is preserved by the kernel but never *forwarded*. All five are now vpushed
  below the `tcr.save_vsp` publish, matching `spentry ffcall` and ppc64's
  `vpush_saveregs`.
* **The pre-`svc` SP step is dropped.** It stepped SP over the whole `c_frame`
  head, which is why every instruction of the old return path read *below* SP
  (`[save3, #c_frame.header]`, `[save3, #c_frame.savedsp]`,
  `[save3, #c_frame.params]`). Its only purpose in the ff-call siblings is to
  place NSAA stack arguments at `[SP]` for the callee, and a Linux/AArch64
  syscall has no stack arguments: `x8` is the syscall number and `x0`-`x5` are
  the at-most-six integer arguments. Leaving SP at the frame head keeps the whole
  `c_frame` above SP for the duration of the trap, and the boundary handed to the
  GC becomes the published boundary `lisp_frame` itself — the address the header
  count strides to — instead of `sp + c_frame.size + 8*node_size`, which is only
  the same address while the frame has exactly eight param words.

Two comments in `spentry ffcall` that described the syscall sibling's old shape
are corrected in the same commit.

### ⚠️ How this was verified — and what the green bar does NOT say

**`.SPsyscall` is unreachable from compiled lisp on arm64.** There is no `.quad`
row for it in the sptab and no `defsubprim .SPsyscall` in `arm64-arch.lisp`, so
nothing compiled can dispatch to it. Nor is anything else wired up to reach it:
the `syscall` vinsn has **no emitter on any port** — `(! syscall)` returns
nothing tree-wide — and there is no lisp-side `syscall` macro or function.

So a full ANSI + `ccl.lsp` run on the rebuilt kernel (**21679/0**, **243/0**) is a
**no-regression** result and **not coverage**. It says the kernel still builds and
boots correctly. It does not say this body ran, because it cannot have. I am
offering this as *correcting a dormant body before anything is wired to it*, not
as a tested fix.

One boundary on that run: it was made before the rebase, on a kernel built from
the earlier base. The patch content did not move. Only its context did.

The mechanism was therefore observed **on the built kernel**, not in the source —
a disassembly lint over the four foreign-transition subprims, asserting that
nothing is reloaded between the transition (`svc`/`blr`) and the
`str xzr,[rcontext,#tcr.valence]`:

| kernel | `_SPsyscall` | the three ff-call bodies |
|---|---|---|
| before this patch | **5 foreign-valence reloads** | clean (fixed in #572) |
| after | clean | clean |

The three ff-call bodies being clean *in the same run* is the control: it shows
the lint can distinguish a fixed body from a broken one, on the same binary, so
the five hits are a finding rather than an artifact of how it reads the file.
The lint is a disassembly check on purpose — the rule is about instruction order
in the shipped binary, and a source grep can be satisfied by text the assembler
reorders or that a later patch re-introduces.

### DIFF
--- a/lisp-kernel/arm64-spentry.s
+++ b/lisp-kernel/arm64-spentry.s
@@ -268,9 +268,9 @@ spentry ffcall
         /* Record the lisp<->foreign boundary for the GC (16m41 protocol,
          * re-pointed for stack args): the boundary is now the PUBLISHED
          * boundary lisp_frame itself -- mark_cstack_area classifies a walk
-         * that STARTS on lisp_frame_marker, exactly as it already does for
-         * the syscall sibling's post-pop boundary -- because the old start
-         * point, the c_frame base, dies once SP steps over it at the blr.
+         * that STARTS on lisp_frame_marker, which is what the syscall
+         * sibling also hands it -- because the old start point, the c_frame
+         * base, dies once SP steps over it at the blr.
          * Everything below the boundary is foreign to the GC while the
          * callee runs, which was already the status of the raw, never-
          * scanned param/stack-arg words when the ivector cover held them.
@@ -292,8 +292,10 @@ spentry ffcall
         str temp0, [rcontext, #tcr.valence]
         /* ARM64-DEVIATION: step SP over header+savedsp+params[0..7] so the
          * callee sees its stack arguments AT [SP], per AAPCS64 5.4.2 (the
-         * single NSAA area the codegen marshals at param words 8..; +80 is
-         * the same c_frame.size + 8 words the syscall sibling steps by).
+         * single NSAA area the codegen marshals at param words 8..; +80 =
+         * c_frame.size + the 8 GPR param words).  The syscall sibling does
+         * NOT do this: `svc' takes no stack arguments, so it leaves SP at
+         * the frame head and its c_frame stays live above SP throughout.
          * PPC64 never moves SP here -- PowerOpen stack params live in the
          * CALLER's frame at positive offsets from the caller's SP; x86-64
          * is the shape donor (_SPffcall's ffcall_setup pops the frame head
@@ -6955,8 +6957,19 @@ endsp callback
  * lisp<->foreign transition) with the AArch64 Linux syscall sequence
  * in the middle instead of a call. */
 spentry syscall
+        /* fn + all four boxed NVRs, exactly as `spentry ffcall'
+         * (arm64-spentry.s -- the CANONICAL NOTE for this whole body), and for
+         * the same reason: the vstack copies are what the GC forwards while
+         * this thread is foreign.  The kernel preserves x19-x22 across the
+         * trap, but preservation is not FORWARDING -- a lisp value that
+         * survives only in an NVR misses any relocation a foreign-era GC
+         * applied.  PPC64 vpush_saveregs()es all eight of its save regs here
+         * (ppc:5404) for exactly this; two of five was an arm64 shortfall. */
         str fn, [vsp, #-node_size]!             /* ppc:5404 vpush_saveregs   */
         str save3, [vsp, #-node_size]!
+        str save2, [vsp, #-node_size]!
+        str save1, [vsp, #-node_size]!
+        str save0, [vsp, #-node_size]!
         mov save3, sp
         /* Park lr in the boundary lisp_frame his alloc-c-frame RESERVED at the
          * frame top, and publish it by shrinking the header count by 4.  The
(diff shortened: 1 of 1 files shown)


xrme:
We haven't used direct syscalls for years, if I recall correctly.  We always go through libc.

Maybe there are some cases still hanging around that I don't know about?

In principle, it's safe to use the syscall interface on Linux, given the kernel's strong commit never to break userland.  On most (all?) other platforms (Darwin, FreeBSD, etc.), the syscall interface is explicitly internal, and you're expected to use the libc interface.

cyclistmass:
You are right, and we checked rather than assuming.

We swept the tree at your current tip. Nothing calls the syscall interface:

- level-0, level-1, lib and library contain no caller.
- No LAP code references `.SPsyscall`.
- No backend `*2.lisp` emits the syscall vinsns, so the x86-32, x86-64 and PPC
  templates are vestigial too.
- Every backend writes `:platform-syscall-mask`. Nothing reads it.
- The C kernel calls only libc's `syscall(2)` wrapper, for futex and gettid.

The one `#_syscall` hit in the tree sits in `#+clisp` code inside asdf.

So the arm64 `.SPsyscall` this patch fixes is dead code that we keep for parity
with x86-64. The patch makes that kept code GC-correct at foreign valence. We
would equally happily delete the spentry instead. Tell us which you prefer and we
will send that version.

One thing we did not establish: the date when the last caller went away. We
worked from a shallow clone.

Separately, this PR and #572 both went CONFLICTING when `503eb61e` deleted
`spentry-E-ffi.s`. We rebased both onto your current tip and pushed them, so they
apply cleanly again. That rewrote the commits, so anything you already read now
has a new hash. The rebase does not change which shape this PR takes, so the
question above still stands.

Mauro DiBenedetto <maurodibenedetto@gmail.com>

xrme:
If you want to bring this patch up-to-date, I'll look at it again.

But given that nothing uses .SPsyscall, I'd rather just delete it unless you've got some reason you want to keep it around.

cyclistmass:
Rebased onto the current tip. One commit now, one file, +73/-36.

#572 has merged, so this collapses exactly as the body predicted, and the
stacked-on-#572 note is gone.

The change itself did not move. What moved under it was `arm64-spentry.s`: the
patch used `msr fpsr, xzr` as context, and that line was deleted by the FPSR
change you took as #605. Context-only refit.

One boundary I have added to the body rather than leave implied. The ANSI and
`ccl.lsp` run quoted there was made before this rebase, on a kernel built from
the earlier base. It was never coverage in the first place, because this body is
unreachable from compiled lisp and cannot have executed. It is a no-regression
result, and that is still what it is.

Issue #587, opened 2026-08-07
Foreign types cannot represent over-aligned composites: wrong sizeof (portable), wrong argument classification (arm64 AAPCS64 C.10, x86-64 SysV)


cyclistmass:
C composites whose alignment is raised above natural alignment
(`_Alignas(16)`, `__attribute__((aligned(N)))`) cannot be described to
CCL's foreign-type system, end to end. Two distinct consequences, one
of them portable.

## Where the information is lost

1. **ffigen5 emits no alignment form.** The `.ffi` grammar already has
   an optional trailing align form on struct/union entries —
   `library/parse-ffi.lisp` reads it into `ffi-struct-alt-alignment-bits`
   (and the union/transparent-union equivalents), and `lib/db-io.lisp`
   encodes it into the `.cdb` — but the libclang-based ffigen5 never
   produces it. (The old gcc-4 ffigen emitted it only for REDUCED
   alignment, the mac68k `#pragma` case.) A grep over a freshly
   generated aarch64 interface corpus finds zero `(align` forms.
2. **The type system cannot represent it.** `parse-field-list`
   (lib/foreign-types.lisp) uses `alt-alignment` only as a per-field
   CAP (`min natural-alignment alt-alignment`), and a record's own
   alignment is the max of its members' natural alignments — which
   never exceeds 8 bytes, since even `long double` is encoded as
   `(:array 2 (:primitive :double))` (lib/db-io.lisp:1329).

So an over-aligned struct is indistinguishable from its
naturally-aligned twin, and no backend can classify it correctly even
in principle.

## Consequence 1 (portable): sizeof is wrong

```c
typedef struct { _Alignas(16) long x; } S;   /* sizeof(S) == 16 */
```

CCL computes the record's size from its fields rounded to the computed
alignment, so `S` is 8 bytes to Lisp. `rlet`/`make-record`
under-allocate; arrays of `S` accessed via `%inc-ptr` use the wrong
stride; a callee that writes through an `S*` clobbers 8 bytes the Lisp
side never allocated. Every platform.

## Consequence 2: argument classification

- **arm64**: AAPCS64 requires a 16-byte-aligned composite passed in
  GPRs to start at an even NGRN (rule C.10; C.8 in older numberings),
  and the NSAA to be rounded up to the argument's natural alignment.
  Without the datum the classification silently disagrees with the C
  compiler about the placement of that argument and of every argument
  after it. The linuxarm64 series at least refuses loudly if the type
[...]

PR #588, opened 2026-08-07
def-foreign-type of a named transparent union fails an ecase

cyclistmass:
Portable bug, found while porting the x86-64 transparent-union
argument handling to arm64, and watched on stock x86-64 CCL 1.12.2:

```
? (def-foreign-type nil (:transparent-union :tu (:p (:* :char))))
> Error: The value :TRANSPARENT-UNION is not of the expected type
>        (MEMBER :STRUCT :UNION :ENUM).
> While executing: CCL::%DEF-AUXILIARY-FOREIGN-TYPES
```

`parse-foreign-record-type` registers a named record as an auxiliary
type under its KIND, and `def-foreign-type`'s expansion replays those
through `%def-auxiliary-foreign-types`, whose ecase knows `:struct`,
`:union` and `:enum` only.  A transparent union belongs to the union
namespace: `parse-foreign-record-type` looks it up with
`info-foreign-type-union`, and the interface-database reader in
db-io.lisp files it there too; the record type's KIND slot keeps the
distinction.

Anonymous transparent unions never hit this (the auxiliary
registration is guarded by the name), which is why databases built by
the old gcc ffigen worked: the reader path does not go through
`def-foreign-type`.

Related: the arm64 argument-passing arm for transparent unions is on
the arm64-ffi-hfa-records branch (#572), and #587 tracks the
over-aligned-composite modeling gap found in the same audit.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/foreign-types.lisp
+++ b/lib/foreign-types.lisp
@@ -576,7 +576,13 @@ Which one name refers to depends on foreign-type-spec in the obvious manner."
 		      (setf (,accessor name) defn))))
 	(ecase kind
 	  (:struct (frob info-foreign-type-struct))
-	  (:union (frob info-foreign-type-union))
+	  ;; A transparent union lives in the union namespace, exactly
+	  ;; where PARSE-FOREIGN-RECORD-TYPE looks it up and where the
+	  ;; interface-database reader files it (db-io.lisp); the record
+	  ;; type's KIND slot keeps the distinction.  Without this arm,
+	  ;; DEF-FOREIGN-TYPE of any named transparent union fails the
+	  ;; ECASE, on every platform.
+	  ((:union :transparent-union) (frob info-foreign-type-union))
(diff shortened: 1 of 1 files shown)


xrme:
Something went wrong here: it looks like your branch is based on arm64. Can you please update that?

xrme (changes requested):
Your branch was cut from arm64, so this shows tons of unrelated commits that I cannot merge.  Please rebase onto master so that only your transparent union fix remains.

cyclistmass:
Rebased onto master. The branch was cut from arm64, which is why the PR carried about 100 unrelated commits. It is now one commit against master, touching lib/foreign-types.lisp only.

The defect is not arm64-specific. We watched it on stock 1.12.2 on x86-64: def-foreign-type of any named transparent union fails the ecase in %def-auxiliary-foreign-types.

PR #591, opened 2026-08-08
parse-ffi: give the reader macro symbols compile-time bindings

cyclistmass:
`compile-file` on `library/parse-ffi.lisp` has never worked, on any platform.

`library/parse-ffi.lisp` reads `*lparen-symbol*` and `*rparen-symbol*` with `#.`
at lines 1194, 1195 and 1332. Read-eval runs at compile time. Plain `defvar`
binds both variables at lines 23 and 24, and `defvar` has no compile-time
effect. The compiler therefore signals `Unbound variable: *LPAREN-SYMBOL*` and
stops.

`load` on the source works. There the `defvar` forms run before the reader
reaches line 1194. `(require "PARSE-FFI")` loads the source, so nothing noticed.

`(defpackage "C" (:use))` on line 22 already has compile-time effect. Wrapping
the two `defvar` forms is therefore sufficient.

Those three sites are the only `#.` in the file. The five neighbouring symbol
`defvar` forms (`*leftbracket-symbol*` and the rest) are never read-evaled, so
this patch leaves them alone.

## Measured

Stock 1.12.2 LinuxX8664:

```
before   ERROR   Unbound variable: *LPAREN-SYMBOL*
                 Read error between positions 52601 and 53298
after    PASS
```

The arm64 branch gives the same failure before and the same pass after. We found
this in a sweep that ran `compile-file` on all 30 `library/*.lisp` files.

## Boundary

We built and ran this on x86-64 and on linuxarm64. We did not build ppc32,
ppc64 or 32-bit ARM, because we have no machine for those ports. The file is
portable and so is the change.

No build list names `library/parse-ffi.lisp`, so no image changes.


xrme:
this change was added in 82e9b1be

PR #592, opened 2026-08-08
Draft: armv7: Add NetBSD support (super WIP turbo draft)

luciusmagn:
This is a very messy and definitely not ready, although pretty much stable (I have been running CCLSH and other things running on top of CCL for a few weeks on this now) port of CCL to NetBSD, first on ARMv7 since that's what I needed.

It is heavily inspired by reading the existing arm 32bit backend and FreeBSD platform as reference

I am just putting it out there so I can start organizing and cleaning this up, there is a lot of work to be done :)

x86 netbsd would probably be the next best target, haha

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ tags
 /*boot*
 
 /armcl
+/narmcl
 /?x86cl
 /?x86cl64
 /wx86cl.exe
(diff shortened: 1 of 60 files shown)


luciusmagn:
I discovered two general bugs that should probably be submitted as their own PR, and merged earlier than this, I will look at it later

PR #593, opened 2026-08-08
arm64: fix two alloc-trap decode defects in pc_luser_xp

cyclistmass:
`pc_luser_xp` has to recognize where a thread stopped inside an inline allocation sequence and either finish or unwind it. Two cases it cannot currently recognize.

**The header-store mask tests the wrong offset.** `IS_SET_ALLOCPTR_HEADER_RD` matches an immediate-9 field of -4, which is `misc_data_offset`. The compiler stores the uvector header at `[allocptr, #-12]`, i.e. minus `fulltag_misc`. The encoding the test looks for is `0xf81fc340`; the encoding emitted is `0xf81f4340`. The mask can never match, so a thread suspended exactly on the header store falls through to `Bug()` instead of finishing the object.

**The pre-trap interior points are unhandled.** The arm64 alloc sequence expands to `sub` / `cmp` / `b.hi` / `udf`, so it has interior points *before* the trap commits. PPC's single trap instruction has no such points, and the port inherited that assumption. ARM32 already solves this in `arm-exceptions.c` by backing the PC up to the `sub` and restarting the allocation; this does the same — rewind to the `sub`, undo its byte accounting if it had executed, and void `allocptr`/`allocbase` so the `b.hi` falls through and the trap re-fires cleanly.

Witnessed by a boot failing with "Unexpected instruction following uvector alloc trap".

**Verification boundary.** Confirmed on Linux arm64 (Graviton4): the baseline kernel fails this 4 times out of 4, the patched kernel is green on the same cell. Not tested on PPC64 or ARM32 — we have no machine for either.

**Scope.** These fix the crash class above. They do **not** fix the nondeterministic boot wedge that was being investigated when they were found; that was tested and did not move. Noting it so the two are not conflated.

### DIFF
--- a/lisp-kernel/arm64-exceptions.c
+++ b/lisp-kernel/arm64-exceptions.c
@@ -274,9 +274,13 @@ void callback_to_lisp(LispObj, ExceptionInformation *, natural, natural,
 #define IS_ALLOC_TRAP(i) ((i) == ALLOC_TRAP_INSTRUCTION)
 
 /* STUR (64-bit): 1111 1000 000 imm9 00 Rn Rt
-   header store: stur Xt, [x26, #misc_header_offset(-4)]
+   header store: stur Xt, [x26, #-fulltag_misc] = [x26, #-12] (16m78:
+     this mask tested -4, which is misc_data_offset, NOT the header
+     offset; the compiler stores the header at P-12, so the old mask
+     could never match and every suspension landing on the header store
+     Bug()d out of finish_allocating_uvector.)
      imm9 = -4 & 0x1ff = 0x1fc → 0xf8000000|0x1fc<<12|26<<5 = 0xf81fc340 */
-#define IS_SET_ALLOCPTR_HEADER_RD(i) (((i) & 0xffffffe0) == 0xf81fc340)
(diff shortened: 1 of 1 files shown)


xrme:
In my bd531b495, I took your fix to IS_SET_ALLOCPTR_HEADER_RD as-is, but I fixed the unhandled pre-trap instructions in the consing sequence in a different way.

Please let me know if I blew it somehow and your boot failure re-arise.

cyclistmass:
Your fix holds. The boot failure did not return.

We advanced our pin to `39bf28ea`, which contains `bd531b495`. At that advance we also dropped our own `0186` from our patch series, because you had absorbed both defects. So the kernel we tested carries your fix and not ours.

The result at that pin:

- ANSI: 21679 tests, 0 failures.
- `ccl.lsp`: 243 tests, 0 failures.
- Exit status 0. The image boots and the full run completes.
- Image md5 `60a4922bffee0568ab62c3b498ec26de`, kernel md5 `64aab3802d957379d9edcc6e497d7fa8`.

The failure was deterministic before the fix. Our baseline failed 4 runs out of 4.

One limit on this evidence. We did not instrument which interior PCs the handler decodes during the run. So we can report that the failure stopped, and that a full suite passes on your version. We cannot report that the run reached every pre-trap interior point you handle.

xrme:
Thank you for the report.

PR #594, opened 2026-08-08
arm64: the #-futex spin-lock release needs RELEASE ordering

cyclistmass:
On linuxarm64 about 1 boot in 40 hangs. The initial thread and the listener both park in `wait_on_semaphore` on the semaphore of the `binding-index-lock` recursive lock (`level-0/l0-symbol.lisp`). Both reach it at boot through `ensure-binding-index`. The wedged lock reads `avail=nwaiters+1, owner=0, count=0, waiting=0`, with every waiter parked and no signaler. That state is unreachable under sequentially consistent execution. The same mechanism also produced occasional "not-lock-owner" errors.

**Cause.** The `#-futex` lock protocols free the lockptr spinlock with a plain store. That store hands off the spinlock, so it must publish every store made while the lock was held. On a weakly-ordered target it needs RELEASE ordering. Without it the next owner reads pre-critical-section values of `avail`, `owner` and `count`. Waiter increments are lost, and an unlock computes `pending=0` while waiters sit on the semaphore. The acquire side is already correct: the Lisp CAS and the C `atomic_swap` both end in `dmb ish`.

This protocol is live on Linux because the feature gate in `level-0/l0-misc.lisp:21` reads `#+(and linux-target no (or x86-target arm-target))`. The `no` comes from `b502cf10` ("work-in-progress") and disables `:futex` for every Linux target on this branch. Stock Linux CCL never runs this code. Darwin always does.

**Evidence, all on one binary** (probe kernel plus runtime redefinition, no image rebuild):

- A contended `ensure-binding-index` hammer reproduces the boot wedge within about 2000 lock cycles.
- Probe kernels logged every exception, suspension and interrupt across a wedge. The fatal window contains no GC, no suspension, no `pc_luser_xp` and no asynchronous signal. The loss is pure concurrent execution.
- The protocol recompiled verbatim hangs 4 of 4 runs. The identical code with a `dmb ish` before each spin-release store runs 10 of 10 clean at 900,000 contended operations per run.

**The fix, in three parts.**

1. `level-0/ARM64/arm64-misc.lisp`: new lap function `%release-spin-lock` (`dmb ish`, then `str xzr`).
2. `level-0/l0-misc.lisp`: all 17 `#-futex` spin-release stores use it. Other targets get an inline plain-store definition, so TSO targets compile as before.
3. `lisp-kernel/threads.h`: `RELEASE_SPINLOCK` on ARM64 calls `release_spin_lock` (`stlr`), which has existed in `arm64-asmutils.s` since the asmutils port but was never wired. The C side shares the lockptr words with Lisp. gcc was also seen sinking `m->waiting = pending` past the plain store, so this stops a compiler reorder as well.

**Verification** of the built image and kernel, Graviton4, quiet box:

- Boot loop: 250 boots, 0 hangs. The unfixed pair measures 1 hang in 220 boots on the same harness.
- Hammer: 10 of 10 runs clean at 900,000 contended operations each. The unfixed pair hangs 4 of 4 on the same harness, run again the same day as a positive control.
- Regression: ANSI 21679 tests, 0 failures. ccl.lsp 243 tests, 0 failures.

Verification boundary: measured on linuxarm64 only. We did not build or test Darwin or ARM32.

**Two observations, reported and not measured.** ARM32 has the same shape: `release_spin_lock` and `atomic_swap_acquire` exist in its asmutils, the C macros do not use them, and its Lisp spin-release stores are plain. Darwin PPC `#-futex` has the same theoretical gap.

**One residual, recorded honestly.** The reproducer wedges only when the `with-lock-grabbed` body executes from the purified readonly section. Byte-identical dynamic-heap copies never trip it, at more than 12M operations, while 12 of 12 readonly runs wedge. Anonymous remap and RWX experiments refuted backing and permission as the variable. We record the placement as an unexplained probability modulator of the race window. The fix removes the wedge in both placements.

### DIFF
--- a/level-0/ARM64/arm64-misc.lisp
+++ b/level-0/ARM64/arm64-misc.lisp
@@ -319,6 +319,28 @@
   (ret))                                           ; ppc:668
 
 ;;; =====================================================================
+;;; %release-spin-lock
+;;; =====================================================================
+;;; Free a spinlock word with RELEASE ordering.  The #-futex lock
+;;; protocols (l0-misc.lisp) publish their critical-section stores by
+;;; freeing the lockptr spinlock; on ARMv8 that hand-off requires a
+;;; release barrier on the freeing store.  With a plain str the next
+;;; spinlock owner can observe pre-critical-section values of
+;;; avail/owner/count -- measured on Graviton4 as a boot deadlock
+;;; (both threads in wait_on_semaphore on the lock's semaphore, frozen
+;;; state avail=nwaiters+1 owner=0 count=0) and as spurious
+;;; not-lock-owner errors.  dmb ish + str is the shape that was
+;;; A/B-confirmed against the reproducer (red 4/4 without, green 10/10
+;;; with, 900k contended ops per run); it also orders the failed-acquire
+;;; waiter path's avail increment before the release.
+(defarm64lapfunction %release-spin-lock ((p arg_z))
+  (macptr-ptr imm0 p)
+  (dmb (:$ 11))                                    ; DMB ISH: release fence
+  (str xzr (:@ imm0))
+  (mov arg_z rnil)
+  (ret))
+
+;;; =====================================================================
 ;;; %fixnum-truncate — ppc-numbers.lisp:251 (#+ppc64-target arm; placed
 ;;; here like called-for-mv-p below, pending the numbers-file story vs
 ;;; Matt's own arm64-numbers.lisp).  Promoted 16m14 from the wave-2
--- a/level-0/l0-misc.lisp
+++ b/level-0/l0-misc.lisp
@@ -565,6 +565,19 @@
       (%atomic-incf-node 1 '*spin-lock-timeouts* target::symbol.vcell)
       (yield))))
 
+;;; Freeing a spinlock publishes every store made while it was held, so
+;;; it needs RELEASE ordering on weakly-ordered targets.  ARM64 defines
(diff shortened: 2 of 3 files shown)


xrme:
This whole area (spinlocks) needs careful review.

https://www.cl.cam.ac.uk/~pes20/cpp/cpp0xmappings.html looks like a possibly-helpful collection of information about mappings of the C/C++11 atomic operations to various architectures.

cyclistmass:
The linux failure was mine. I fixed it.

The `#-futex` arm declaimed `%release-spin-lock` at top level. In level-0 a
top-level `declaim` compiles a load-time `proclaim` call into the boot image,
and `proclaim` is level-1.

The fix wraps it in `eval-when`, which is this file's own idiom a few lines
below. I reproduced the failure on x86-64 first. I should have built x86-64
before I sent this.

On your review point, the change only adds ordering:

- Lisp release: `dmb ish` then `str`. That is the leading-fence mapping in your
  table.
- C release: `stlr`, through `release_spin_lock` in arm64-asmutils.s.
- Acquire: unchanged. Both sides already end in `dmb ish`.

The two release sides use different mappings, which deserves your eye. I used
`dmb ish` plus `str` because that is the shape I confirmed by A/B. `stlr` should
be equally correct. I did not test it.

arm64 after the fix: ANSI 21679/0, ccl.lsp 243/0, and the six patched lock
functions disassemble identically before and after.

PR #596, opened 2026-08-09
arm64: combine the spentry-[A-E] files into arm64-spentry.s

cyclistmass:
You said this needed doing. I am sending it without waiting for your reply, in
case it saves you the time. If you have already started, drop this.

The five files become one `arm64-spentry.s`, in the old SPOBJ link order, with
the sptab last. That keeps the linked `.text` layout as it was.

The merge forced three changes:

- One include block instead of five.
- One copy of the five duplicated macros. The bodies match.
- The misc_ref sketch's file-local `C(misc_ref_common)` becomes
  `misc_ref_sketch_common`, because cluster B carries the real one. That is the
  only name I changed.

`lisp-kernel/linuxarm64/Makefile` drops the five objects from SPOBJ.

The sptab is the part that fails silently, so I checked it twice: 133 of 133
rows identical in the source, and 133 of 133 in the linked kernel. I broke each
check first and watched it catch the break.

Built and run on linuxarm64 at 18873bd1. ANSI 21679/0, ccl.lsp 243/0, exit 0,
identical to the pre-combine baseline.


xrme:
I already was working on this, so I committed my version 503eb61e.

Issue #13, opened 2017-02-18
Update or replace ffigen4


xrme:
The interface databases that CCL uses are generated by a program called ffigen4. It is a set of patches to gcc-4.0.0 (see http://svn.clozure.com/publicsvn/ffigen4/)

These patches should be brought up-to-date. Alternatively, it might be an option to replace ffigen4 with some other tool.  https://github.com/rpav/c2ffi might be suitable.

eschaton:
What about just using CFFI? Or would that be insufficient?

xrme:
The interface databases make the `#_` and `#$` reader macros work.  These reader macros are used extensively in the implementation of CCL itself.

I'm not a CFFI user, so I'm not really qualified to say whether it is nicer than CCL's native FFI, but I can say that I think that CCL's native FFI is a great feature.

ailisp:
I know that CFFI is only a portable layer. Like what bordeaux-threads do with CCL's multiprocessing. So it may be not appropriate to use CFFI here,since we don't need to use the interface database in other CL and also Clozure's FFI provide more functionality. I suggest to update ffigen, because writing a new backend for c2ffi may only interested for developing CCL itself, library authors usually use CFFI other than platform specific ones.

Linn0:
I agree that CCL's native FFI is a great feature. But unfortunately, rarely projects build upon it, instead they actually build upon CFFI. so I made a little project ccl-cffi to use same function interface as CCL, but implement it upon CCL more efficiently.

ailisp:
I would like to work on this. Today I took a look at ffigen, and CCL's ffi doc. Looks not good to always patch gcc to build ffigen. Using https://github.com/rpav/c2ffi  is a good idea. Here I can either
1. add a driver for c2ffi to generate ffi format
or
2. replace lib/parse-ffi.lisp with a new lisp program that input c2ffi's sexp output and output cdb.
@xrme Which way you prefer? Approach 2 is using lisp, of course more fun than c++ :) Thanks!

xrme:
You are welcome to work on this if you want to, but I worry that it is a rather big project.

I agree that we should try out https://github.com/rpav/c2ffi.  If we can parse a simple header file with c2ffi and convince ourselves that the output matches (well, is isomorphic to) the current ffigen output, then that will give us some confidence that we can make it work.

[...]

PR #598, opened 2026-08-11
ffi: an array-typed function parameter must decay to a pointer

cyclistmass:
An array-typed function parameter reaches the interface database as `#\?`, the "unknown" marker. Nothing decodes that marker. The argument spec becomes NIL, and a caller then fails at macroexpansion with `Unknown foreign type: NIL`. The error names neither the function nor the array.

This is not arm64. I measured it on stock 1.12.2, LinuxX8664:

```
(encode-ffi-arg-type '(:array 20 (:primitive (:unsigned 8))))  =>  (#\?)
compiling a call through such a definition  =>  Error: Unknown foreign type: NIL
```

Older ffigen decayed arrays before it wrote the database, so no committed .ffi file holds an array parameter. I scanned our 82 and found none. clang 19.1.7 emits them verbatim. This may bear on #13.

The fix encodes an array parameter as a pointer, which C requires (C11 6.7.6.3p7). The rest of the chain agrees already. `foreign-type-to-representation-type` maps `foreign-array-type` to `:address`. The ppc64 and arm64 callback generators read such an argument with `%get-ptr`. The second hunk gives `translate-foreign-arg-type` the same arm.

`encode-ffi-type` keeps full array types for typedefs and record fields, and I left it alone. An array must not decay there.

I watched the test fail, then pass, on one variable. It ends in a live `tmpnam(3)` call that fills a 20-byte buffer.

`lib/db-io.lisp` matches byte for byte on master, so master carries the same defect. I based this on arm64 because `lib/foreign-types.lisp` differs and the second hunk needs the arm64 text.

I have no PowerPC machine. I did not run this on ppc64.

### DIFF
--- a/lib/db-io.lisp
+++ b/lib/db-io.lisp
@@ -1442,8 +1442,12 @@ satisfy the optional predicate PREDICATE."
      `(#\t ,@(encode-name (ffi-typedef-name (cadr spec)))))
     (:pointer
       `(#\a))
+    ;; C adjusts a parameter of array type to a pointer to the element
+    ;; type (C11 6.7.6.3p7); clang-era ffigen emits the array type
+    ;; verbatim, so perform the adjustment here.  Arrays do not decay
+    ;; in ENCODE-FFI-TYPE (typedefs, record fields), and must not.
     (:array
-      `(#\?))))
+      `(#\a))))
 
 (defun encode-ffi-arg-list (args)
   (if args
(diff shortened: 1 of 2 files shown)


cyclistmass:
I cited 1.12.2 above. The same result holds on stock 1.13, so this is not a defect that a newer release already handles:

```
Version 1.13 (v1.13-309-g7cff8ca5) LinuxX8664

(encode-ffi-arg-type '(:array 20 (:primitive (:unsigned 8))))  =>  (#\?)
(encode-ffi-arg-type '(:pointer (:primitive (:unsigned 8))))   =>  (#\a)
(decode-arguments "a?")                                        =>  (NIL), :ADDRESS
```

The pointer case encodes correctly in the same image. Only the array arm writes the unknown marker.

PR #599, opened 2026-08-13
darwinarm64: native Apple Silicon support — Mach kernel, W^X, Darwin AAPCS64 policy, ObjC bridge, Cocoa IDE

egao1980:
## Summary

Disclaimer: This is AI coded, human guided and tested work. 

Native Darwin/arm64 (Apple Silicon) support on top of the current `arm64` branch, in seven commits:

1. **native kernel** — `lisp-kernel/darwinarm64/` builds `darm64cl`; Mach exception server (`arm64-darwin-mach.c`, ported from the darwinx8664 path); MAP_JIT-based W^X (executable lisp code in a MAP_JIT mapping, purify moves pure objects to `AREA_READONLY` RX, dynamic heap never executable, write-protect toggles only in kernel C); GC corrupt-header diagnostics; 16KiB-page-aware guard stepping.
2. **runtime and FFI** — the OS-independent AAPCS64 logic (classification, `expand-ff-call`, callback glue) moves from `lib/ffi-linuxarm64.lisp` to a shared `lib/ffi-arm64.lisp`; both OS files are now thin delegators. Darwin ABI divergences ride on top, keyed on the target OS: variadic arguments are stack-only behind a `:variadic` sentinel (with the objc_msgSend family excluded — its `...` prototype does not use the C variadic convention on arm64), and non-variadic stack overflow packs scalars at natural size/alignment. Also: Darwin syscall convention (x16, `svc #0x80`, carry→-errno), `bcond_ext`/ADRP/`C()` assembler portability for Mach-O, save0–save2 GC-visibility spills in `.SPsyscall` (same contract as ffcall), `%throw`/`%throwing-through-cleanup-p` for `:propagate-throw` callbacks, and MAP_JIT-safe callback trampoline installation.
3. **ObjC bridge and Cocoa IDE** — Apple tagged pointers, no `objc_msgSend*_stret` on arm64, varargs sends lowered onto the interpreted `%ff-call`, `call-next-method` over ObjC methods, untitled Listener at launch, Trace/Source/Inspect selection fixes. `(require :cocoa-application)` builds Clozure CL64.app end to end.
4. **interface database tooling** — scripts to regenerate the darwin-arm64 `.cdb` databases from the current SDK with ffigen5; the generated databases stay out of the tree.
5. **stock build path** — `(rebuild-ccl :full t)` works like other platforms (cross xload → cold load → purify → save-application); dumplisp clears MAP_JIT macptrs before the image is written.
6. **smoke tests and CI harness** — FFI/callback/Cocoa smokes, a byte-stable launch-layout regression reproducer for the ffcall GC-visibility bug class, Cocoa mini-apps, and `tools/run-darwin-arm64-ci.sh` as a single gate.
7. **porting notes** — `doc/porting/darwin.md` (why `-pagezero_size`/`-no_pie` don't exist on arm64 Darwin, the W^X policy and how V8/JSC handle MAP_JIT), `doc/porting/progress.md`.

Where this branch overlapped work that already landed on `arm64` (#572's AAPCS64 FFI, the spentry consolidation, the x10 callback-index trampoline, interpreted `%ff-call`), the in-tree implementation was kept and only the Darwin deltas were added.

## Validation (Apple Silicon, macOS)

- `make -C lisp-kernel/darwinarm64` → native `darm64cl`
- `(rebuild-ccl :full t)` end to end, plus a forced full self-host rebuild (`:force t`) from the produced image
- ANSI + CCL suites from ccl-tests: **all 21920 tests pass** (~40s)
- Full smoke gate (`tools/run-darwin-arm64-ci.sh`): math, purify save/reload, cocoa CDB/bridge, interpreted ff-call, clean build, `(require :cocoa)`/Hemlock, AppKit mini-apps
- FFI/callback smokes: struct return (HFA NSSize/NSPoint/NSRect, GPR NSRange), callback sret/struct/HFA (incl. runtime `drawRect:`), HFA arguments, variadic (`#/stringWithFormat:`), natural-size pack overflow, NSRange foreign-type resolution, ffcall stack discipline, ObjC bridge, `call-next-method`, init keywords
- Launch-layout GC regression reproducer passes; the stock-built IDE opens its untitled Listener and stays healthy
- Deprecated AppKit names (`NSOffState`, `NSCommandKeyMask`, `NSOKButton`, `NSNotFound`, window masks, bezel styles, control sizes, ...) are `static const` aliases in modern SDKs, so freshly translated interface databases record them as foreign variables that fail to resolve at load time (the `NSOffState` error seen above). The IDE, Hemlock bindings, and the bundled easygui/progress examples now use fixed-ABI Lisp constants (`cocoa-ide/constants.lisp`) instead — audited against every `#$` name the IDE build references, so bring-up no longer depends on any of those names being present in the `.cdb`s.
- Verified end-to-end against databases regenerated with the ccl-ffigen#6 `translate.sh` flow on the current SDK: with this branch's injectors applied (`inject-objc-msgsend-prototypes` — modern `objc/message.h` declares `objc_msgSend` argless, without it `(require :objc-support)` fails with "Extra arguments"; `inject-objc-bridge-types`; `inject-objc-protocol-class`; and the new `inject-libc-computed-constants` for `sizeof`/cast macros the translator cannot reduce: `FIONBIO`, `HOST_BASIC_INFO_COUNT`, `CPU_TYPE_*`), the full rebuild, ANSI+CCL suites (21920/21920), smoke gate, IDE app build, and IDE interaction smokes all pass on both the regenerated and the previously generated databases. The carbon/gl/quartz/quartzcore databases come out byte-identical between the two flows.
- Cocoa callback error handling: a Lisp error inside an ObjC callback (e.g. a table-view action) previously corrupted the error-return trampoline (it wrote MAP_JIT memory directly) and could wedge the event thread; the trampoline is now installed via the kernel blit path, the xapropos row -1 action is guarded, and a click-regression smoke covers the flow

Linux/arm64 is intended to be unaffected: the shared `lib/ffi-arm64.lisp` is a file move of the linuxarm64 implementation (the Darwin policies are dormant behind target-OS checks), and the kernel assembly changes are `#ifdef`-gated or expand to the previous instructions on Linux (`bcond_ext`, `load_addr_of_lisp_nil`). I could not run a linuxarm64 build here; happy to fix anything a Linux CI run turns up.


gmpalter:
@xrme is away until next Tuesday, the 18th.

egao1980:
I'm still working on IDE parts so it definitely can wait.

egao1980:
IDE looks fine at this point including inspector.

gmpalter:
I just built a macOS Arm64 CCL.

I'm not certain that the CDB files are all ok, however. I'm going to submit a PR to the ccl-ffigen repository with what I did based on your work. (I've replaced the XXX-populate.sh scripts with appropriate translate.sh scripts.)

gmpalter:
Try to bring up the IDE gets the error ...
```
> Error: Can't resolve foreign symbol "NSOffState"
> While executing: CCL::RESOLVE-FOREIGN-VARIABLE, in process Initial(0).
```

gmpalter:
Here's the ccl-ffigen pull request -- https://github.com/Clozure/ccl-ffigen/pull/6

egao1980:
@gmpalter I've pushed few changes and fixes - your PR is looking great, you probably can add ObjC injectors / libc one from this branch too (inject-* files).

gmpalter:
You don't need inject-libc-computed-constants.lisp.

The trick with parse-standard-ffi-files is to run it twice. The second run will resolve those computed constants.

gmpalter:
The problem I reported above about `NSOffState` is either an ffigen5 or parse-ffi problem.

In `AppKit.framework/Headers/NSCell.h`, we find these two definitions ...
```
static const NSControlStateValue NSControlStateValueOff = 0;                                   
static const NSControlStateValue NSOffState API_DEPRECATED_WITH_REPLACEMENT("NSControlStateValueOff", macos(10.0,10.14)) = NSControlStateValueOff;                                           
```
Those should have resulted in a definition in the CDB for `NSOffState` but didn't.

gmpalter:
Here's a warning ...

I updated from macOS 26.6.1 to 26.6.2 yesterday. Now, when I try to launch the image I built a few days ago, I get
```
[palter@miniMe ~/CCL/ccl-macosarm64](15)$ ./darm64cl
Couldn't load lisp heap image from /Users/palter/CCL/ccl-macosarm64/darm64cl.image: Invalid argument
```
I'm going to try cross compiling and building a new image to see if that'll work.

gmpalter:
I just did a fresh cross-compile and cross-xload-level-0. I recompiled the kernel. Same error. 26.6.2 has broken this port!

```
[palter@miniMe ~/CCL/ccl-macosarm64](38)$ ./darm64cl -I arm64-boot.image 
Couldn't load lisp heap image from arm64-boot.image: Invalid argument
```

egao1980:
I'll check a bit later today.  Thanks for the feedback.

gmpalter:
It looks like the first call to `mmap` in `CommitMemory` in images.c is failing. The call is
```
   215 	    /* W^X: RWX mmap is rejected.  Dynamic heap is RW only; executable
   216 	       code lives in MAP_JIT (AREA_CODE) or AREA_READONLY after purify. */
-> 217 	    addr = mmap(start, len, MEMPROTECT_RW, MAP_PRIVATE|MAP_ANON|MAP_FIXED, -1, 0);
```
with `start = 0x000000020000000` and `len = 16384`.

I've attached a sample from Activity Monitor which shows where the various libraries are loaded in memory. (I'm not sure why `libRosetta` is loaded as this is definitely native Arm code.) Although nothing's loaded into those addresses, it does fall between a couple of libraries. I wonder if macOS is now refusing to map anything between libraries.

[Sample of darm64cl.txt](https://github.com/user-attachments/files/31190176/Sample.of.darm64cl.txt)

gmpalter:
For the record, the Symbolics Virtual Lisp Machine (VLM) which I maintain also maps its data into fixed locations in the address space and it runs without problems. The address ranges it uses are 0x10000000000 - 0x100FFFFFFFF and 0x40000000000 - 0x4003FFFFFFF.

Either you'll need to change where areas are mapped into memory or devise a way to make areas relocatable. I suspect the former is easier but has the problem that future macOS updates could subsume the address range requiring finding a new range. The latter is more difficult as I presume there are pointers inside the areas that would need to be relocated. (Perhaps some of the GC's mechanisms could be used to do that.)

egao1980:
@gmpalter I think I've found (and fixed) your issue. Please check.

gmpalter:
I just tried your latest commits and was able to build and run CCL on macOS 26.6.2, thank you.

I presume you'll remove the message
```
note: static space at 0x102628000 (canonical 0x200000000 unavailable: Permission denied)
```
before this is merged into the arm64 branch.

gmpalter:
@xrme has merged my PR for generating the darwin-arm64-headers into https://github.com/Clozure/ccl-ffigen. So, you no longer need your `tools/darwin-arm64-cdb` directory.

xrme:
I've incorporated the changes that are needed to make the tty lisp work. Thank you very much.

They've landed on the arm64 branch.  I'll next work on reviewing and bringing over the Cocoa-related changes.

One thing I think that is worth noting: the MAP_JIT region as currently implemented is fine as an interim mechanism to get the ball rolling, but for the darwinarm64 port to be real, a more robust scheme will be needed.  (I don't think I'm saying anything you don't already know.). But having a native, self-hosting CCL on darwinarm64 is a win, even if there's more to do.  Again, thank you.

xrme:
I brought in (to the arm64 branch) most of the objc bridge and cocoa IDE files now, too.

The next thing to verify is to freshly rebuild the interfaces with ccl-ffigen and process them with the parse-ffi as found on the arm64 branch.

I omitted your special-casing of static const that you have in c9b128f2; it may be that I need to pick that up when I get around to trying a fresh rebuild of the interfaces.

Like I mentioned in the commit messages, it should now work to do (require 'cocoa) from the arm64 branch.

no-defun-allowed:
> MAP_JIT-based W^X (executable lisp code in a MAP_JIT mapping, purify moves pure objects to AREA_READONLY RX, dynamic heap never executable, write-protect toggles only in kernel C)

I did most of this already in my Rosetta-friendly x86-64 port <https://github.com/no-defun-allowed/ccl/blob/master/lisp-kernel/code-gc.c> and <https://github.com/no-defun-allowed/ccl/blob/master/lisp-kernel/image.c#L336-L350>. Thank you for nothing, you sloppers.

xrme:
> I did most of this already in my Rosetta-friendly x86-64 port https://github.com/no-defun-allowed/ccl/blob/master/lisp-kernel/code-gc.c and https://github.com/no-defun-allowed/ccl/blob/master/lisp-kernel/image.c#L336-L350. Thank you for nothing, you sloppers.

@no-defun-allowed I expect that the interim implementation currently here will be superseded by something based on what you already did.  I know not everyone is going to think it's a good idea to accept contributions like this, but I believe that having a working native CCL on macOS/arm64, even if it is hackish in ways, is worthwhile and will enable faster future progress.

egao1980:
> > MAP_JIT-based W^X (executable lisp code in a MAP_JIT mapping, purify moves pure objects to AREA_READONLY RX, dynamic heap never executable, write-protect toggles only in kernel C)
> 
> I did most of this already in my Rosetta-friendly x86-64 port https://github.com/no-defun-allowed/ccl/blob/master/lisp-kernel/code-gc.c and https://github.com/no-defun-allowed/ccl/blob/master/lisp-kernel/image.c#L336-L350. Thank you for nothing, you sloppers.

Hi @no-defun-allowed, I would also prefer human written / reviewed code. That's why this PR is marked as AI generated human tested draft. This is an expensive web search / prior art study / make it work effort. I've created quite a few smoke tests to help porting CCL to M chips in case the changes won't end up in the main project. 
Common Lisp is currently a niche language and PRs can sit in OSS project repos for years (ask me how I know that :). CCL port is a tough challenge - you need either a team or a senior developer with CL + arm64 assembly + MacOS system internals + Cocoa skill set and lots of free time.  
Please treat this code as a kickstart to get a working version with IDE until we have the proper release. Please please keep working on CL projects.

mctuva:
Dear @no-defun-allowed, @egao1980 and others who are generally disgusted by LLM-written code,

I feel like I owe you an apology for the ClaudeCode Fable-generated, human-overseen port of CCL to the ARM64. Honestly, I really, really didn't think this effort would succeed, but my colleague Mauro was so enthusiastic about finishing the project, that I felt obliged to offer him first encouragement and then kudos.  Yes, I can imagine there is no shortage of ugly, unmaintainable code that RME and friends will want to clean up (maybe you?), but the fact is that, not only does this pass the CL Test Suite, but our tiny software company got one of its most complex, commercial Lisp applications up and running on this ARM port, providing further validation that this is not a toy.

For bit of extra perspective, I have been trying to recruit and get funding for a human-written ARM64 port for the last 5 years, and either money, free time, or expertise was always in short supply.  Despite that, here we are.  We have an LLM-ported CCL that works.  Maybe it is best to think of it as a starting point for a better, cleaner, more maintainable port (and please try not to be too grossed out by what you see).  Remember, it's a lot easier to beautify something that works already, than to know you have months of tireless work ahead to simply get to "hello world."

Meanwhile, three cheers to brilliant, longtime kernel and compiler author and general CCL maintainer, Gary Byers, who laid the real groundwork to make this port at all possible!  The LLM started with his PPC64 implementation, which is similar enough to the ARM64 in so many ways.  But more importantly, many parts of the kernel were designed to be portable, and the legacy of the original 68k Mac implementation, plus the PPC32 and 64, ARM32, Intel32 and 64 ports really did make it more likely for ClaudeCode to succeed (as much as we might have, at best, a love-hate relationship with it and its "glorified auto-complete" and robber of other people's work and words).  

--Tim (aka McTuva)

Issue #530, opened 2025-04-04
Can't read #.(values ...)


stassats:
```
Clozure Common Lisp Version 1.13 (v1.13-20-gcc834499) LinuxX8664

? (funcall #.(compile nil `(lambda () )))

> Error: (#<Anonymous Function #x3020004D85EF> NIL NIL) can't be destructured against the lambda list (CCL::FORM &OPTIONAL CCL::NESTED-SOURCE-NOTES), because it contains 3 elements, and at most 2 are expected.
> While executing: CCL::PREPARE-TO-DESTRUCTURE, in process listener(1).

```

stassats:
Or just (list #.(values 1 2 3))

PR #607, opened 2026-08-16
fix: `#.` (read-eval) reader-macro crash when eval `#.(values ...)`

iwami4438:
- Fixed the code because it simply used `eval` and did not anticipate the use of `values`, which returns multiple values.
- Ref: https://github.com/Clozure/ccl/issues/530


xrme:
Thank you for tracking down the location of the problem. 

I ended up fixing it in another way in 46a05103.

I also added a couple of tests in https://github.com/Clozure/ccl-tests/commit/c4bc7e9a1c453d677353932362e30b589e2c2e40

PR #609, opened 2026-08-19
arm64: populate the NVR register pool, and the four changes it requires

cyclistmass:
This branch is the arm64 NVR register pool: seven commits, one subject.
Everything unrelated has been removed, as agreed in the comments.

`*arm642-nvrs*` is nil (arm642.lisp:206), so every lexical is homed on the
value stack. Patch 1 populates it with save0..save3. Details per patch are in
the commit messages.

**Aggregate: 10.5% less DO-TESTS CPU time** — 49.86 s against 55.72 s median,
6 interleaved pairs on one Graviton5, pool-on wins all 6, ranges do not
overlap. Hardware cycles agree at 10.2% fewer. The pool executes 3.9% MORE
instructions and still wins, IPC 3.041 to 3.520, so an instruction-count
argument predicts the wrong sign. ANSI 21679/0 and ccl.lsp 243/0 on all 12
runs.

The declared-array inner loop goes from 6.98 to 2.98 cycles per element.
Costs: deep recursion 1.16x slower, small allocation 1.11x to 1.13x, both from
save and restore on every call. The aggregate is the net.

1. Populate the pool.
2. Restore NVRs from a pre-spread vsp snapshot. **A correctness fix you can
   take alone.** A tail spread moves vsp below the save area, so the restore
   reads pushed arguments into save0-3 and they ride the return chain into
   every caller. Latent while the pool is empty. PPC64 donor,
   ppc2.lisp:2385-2391.
3. Float-typed lexicals never win a GPR NVR. Ports x862.lisp:7297-7305.
4. and 5. **Do not take patch 1 without these.** With the pool live a
   pool-homed local reads :UNAVAILABLE in the debugger. Patch 4 encodes the
   unused `:regsave` note into a code-vector trailer and carries a test we
   watched fail first; patch 5 supplies the portable count and names.

Not tested: Apple hardware, rebuild-ccl.

### DIFF
--- a/compiler/ARM64/arm64-lap.lisp
+++ b/compiler/ARM64/arm64-lap.lisp
@@ -84,7 +84,13 @@
       (arm64-lap-generate-code current section-size *arm64-lap-lfun-bits*)
       )))
 
-(defun arm64-lap-generate-code (seg code-vector-size &optional (lfbits 0))
+;;; TRAILER-WORD, when non-nil, is appended as one extra data word at
+;;; the very end of the code vector (the register-save trailer --
+;;; format and rationale at arm642.lisp arm642-encode-regsave-trailer).
+;;; Hand-written lap functions never pass it: they manage save0-3
+;;; explicitly and record nothing.
+(defun arm64-lap-generate-code (seg code-vector-size &optional (lfbits 0)
+                                    trailer-word)
   (declare (fixnum code-vector-size))
   (let* ((target-backend *target-backend*)
          (cross-compiling (target-arch-case
@@ -101,7 +107,8 @@
          (i prefix-size))
     (declare (fixnum i constants-size))
     (let* ((code-vector (%alloc-misc
-                         (+ code-vector-size prefix-size)
+                         (+ code-vector-size prefix-size
+                            (if trailer-word 1 0))
                          (if cross-compiling
                            target::subtag-xcode-vector
                            arm64::subtag-code-vector))))
@@ -111,6 +118,9 @@
         (unless (eql (arm64::instruction-element-size insn) 0)
           (setf (uvref code-vector i) (arm64::instruction-word insn))
           (incf i)))
+      (when trailer-word
+        (setf (uvref code-vector i) trailer-word)
+        (incf i))
       (dolist (pair arm64::*constants*)
         (let ((imm (car pair))
               (k (cdr pair)))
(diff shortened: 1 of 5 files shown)


xrme:
I suppose you made this to keep everything together in one place, but it's awfully large for me to possibly merge all at once.

cyclistmass:
You are right, and I am sorry — that PR should not have been opened in that shape.

Size is the visible problem. The real one is duplication: **13 of its 25 commits are changes you already have**, in nine open PRs (#575, #576, #577, #578, #590, #598, #602, #604, #605). I sent you the same diffs twice and left you to reconcile a bundle against your own review queue.

What I will do unless you prefer otherwise:

- Reduce #609 to the NVR register pool alone: 7 commits, one subject, no overlap with anything already open.
- Take the 5 unrelated fixes out and hold them, rather than open five more PRs while you are holding ten.

Separately, and worth your time: the failing check on #609 is not caused by its contents. The single failure is `CL-TEST::SHARPSIGN-DOT.VALUES.MULTIPLE`, which your `46a05103` fixes — the branch predates it. The green checks on the PRs above all ran between 11 and 14 August, before that test could fail, so they would fail the same way if re-run on a stale base.

Tell me what order you want these in and I will follow it.

xrme:
Let me resolve the already-open PRs.  It's fine to narrow #609 as you propose.

cyclistmass:
Narrowed as agreed. Seven commits, the NVR register pool only:

```
arm642: restore the NVRs after a spread only for a TAIL call
arm642: populate the nvr register pool
arm642: restore nvrs from a pre-spread vsp snapshot
arm642: float-typed vars never win a gpr nvr
arm64:  backtrace recovers NVR-homed variables (register-save trailer)
lib:    arm64 saved-register count and names for the populated pool
ppc2:   float-typed lexicals must never win a GPR NVR
```

Five files, +270/-39, and it now reports mergeable.

Everything else is out. Ten commits are already in your tree from the PRs you
merged. `.SPsyscall` stays in #578. The cs-limit probe is dropped outright rather
than held, because your `3d4c133b` supersedes it. The remaining unrelated fixes I
am holding rather than opening more PRs.

This was force-pushed, so the earlier SHAs are gone.

xrme:
This landed on arm64 as four commits (714e8604a..104d13651), listing you as co-author.

The register-save metadata lives in the function object's %lfun-info plist (packed into a fixnum under regsister-save-info) instead of a code-vector trailer. Same info, but it keeps the non-instruction data out of code-vectors.

I dropped the ppc2.lisp float-NVR guard. It's probably right, but since we can't build or test a PPC Lisp, I'd rather not merge untested code to that port.

The observed ~10% less CPU running the tests is a nice result.

PR #614, opened 2026-08-22
arm64: a derived code PC must not rest where the GC cannot relocate it

cyclistmass:
Four commits on `d61da8d0`. Commits 1 to 3 are one defect class. Commit 4 is
unrelated. Each commit message carries its own detail.

Your `df738b0e` states the class better than my patch text did. *"Keeping a
pc-locative in the node register temp4 is not gc-safe."* These are the
remaining sites.

**The class.** `mark_xp`, `forward_xp`, `purify_xp` and `impurify_xp` treat
`xpPC` and `xpLR` as pc-locatives, and `x7`..`x23` as NODES. An untagged code
address in a node register therefore reads as a fixnum and stays unrelocated.
The GC never scans `x0`-`x6`. A thread crashed branching to an unwind-protect
cleanup at the PRE-GC address of a codevector a compacting GC had moved.

Every other port closes this deliberately. PPC keeps such PCs in `loc_pc`,
x86-64 tags them `tra`, and ARM32 keeps them in `lr` with one 3-instruction swap
window that `pc_luser_xp` rolls forward. arm64 took PPC's idiom onto
`temp4`/`x16` without its GC half.

**Commit 1 adopts the ARM32 model,** because neither `tra` nor a spare register
exists here. `mkcatch` stores `lr` itself as the catch frame's `savelr`, so the
walker relocates that slot. The 7-instruction derive window is gone. The
`nthrow` paths hold the cleanup PC in the frame slot until a 4-instruction
exported window moves it to `lr`, then call `blr lr`. That window also zeroes
`temp4`, so the GC never node-scans a dead interior PC.

Your `df738b0e` supersedes four of its hunks and I dropped them. Your `nvalret`
form is stronger than mine: fall-through, label deleted.

**Commit 2 fixes four frame builds `pc_luser_xp` cannot see.** Case (a)
recognizes a `str` to `sp` only when the frame-creating `sub` sits 1 to 3
instructions back. `mvpass`, `mvpasssym` and `lexpr_entry` use a `sub` and four
separate `str` instructions, which puts `savefn` and `savelr` 4 and 5
instructions out. A thread suspended there shows a valid marker, a garbage
`savefn` which the GC node-scans, and a garbage `savelr` which it
locative-scans.

All four now use the `stp` pair, and case (a2) zeroes both slots for a thread
caught between the two `stp`s. Case (a2) also had to widen: it pinned `Rt2` to
`vsp`, but three of the four push a computed entry-vsp from `imm0`. It now frees
`Rt2` and confirms the stored marker word instead. The push has already executed
at the recognized PC, so a real frame build always passes and a coincidental
`stp` pair cannot.

**Commit 3 is the entry-path sibling of your `df738b0e`.** Every lexpr and
`&rest` prologue ended `mov temp4, lr` and called `_SPlexpr_entry`, which stored
`temp4` to FRAME-A `savelr` eight instructions later. The `*gf-proto*` and
`*cm-proto*` LAP prototypes carried the same window. The fix moves FRAME-A into
the callers, which build it with commit 2's shape and store `LR` to `savelr`
before the call.

`pc_luser_xp` cannot close this window, by construction. It rolls forward the
partially built state it recognizes, and it normalizes the allocation
registers. It does not relocate a general register, and no case in it reads or
rewrites `temp4`.

Your shape does not transfer here, for one reason. `values` and `nvalret` hold
one live pc and call nothing, so it can stay in `lr`. The lexpr prologue holds
two live pcs across a call, and one `lr` cannot hold both, so the second home
must be the walked slot. This is your design under that constraint, not a second
design.

**Commit 4 is memory ordering, not GC.** The arm64 LAP atomics came from PPC64
instruction by instruction, and PPC has no acquire or release load or store.
Three sites therefore still pay a full `dmb` for ordering that ARMv8 already
provides. `%atomic-incf-node` and `%ptr-store-fixnum-conditional` become
`ldaxr`/`stlxr` with no fence, and `%release-spin-lock` becomes one `stlr`.
Every instruction is baseline ARMv8.0. ARMv7 has no `stlr`, so the ARM32 idiom
stays `dmb` plus `str` and this does not carry back.

Measured per operation on a 2-core arm64 Linux host: uncontended
`with-lock-grabbed` -18%, two threads on two locks -19%, `atomic-incf` -10%.

**Verified at `d61da8d0`.** The four commits chain-apply clean on a pristine
tree. Built and tested on linuxarm64 at that base: ANSI `:TOTAL 21679 :FAILED 0
:EXCLUDED NIL` and CCL `:TOTAL 243 :FAILED 0`.

**Scope.** Every file is arm64-scoped: `lisp-kernel/arm64-spentry.s`,
`lisp-kernel/arm64-exceptions.c`, `level-0/ARM64/arm64-clos.lisp` and
`level-0/ARM64/arm64-misc.lisp`. I have not built any other target.

**Two limits.** No red test forces a suspension inside these windows. Each is a
few instructions wide, and only another thread can suspend inside one. Both
suites are also single-thread, so they cannot certify a barrier-ordering change.

For commit 4 the ordering evidence is a litmus suite of 5 families and 12
variants. All 6 red controls fired. 155,435,310 synchronized iterations gave
zero anomalies, and every variant agrees with stock x86-64 CCL. That shows the
change is safe. It does not show it is necessary, because the `dmb` form is
strictly stronger and also passes.

`0215`, the `pc_luser_xp` EGC dispatch order, is deliberately not here. It is
already PR #611 and applies independently.

### DIFF
--- a/level-0/ARM64/arm64-clos.lisp
+++ b/level-0/ARM64/arm64-clos.lisp
@@ -242,11 +242,17 @@
      (arm64-lap-function
       gag
       ()
-      (mov temp4 lr)                        ; ppc:205 (mflr loc-pc) — LEXPR-RA
       (vpush-argregs)                       ; ppc:206
       (vpush nargs)                         ; ppc:207
       (add imm0 vsp nargs)                  ; ppc:208
       (add imm0 imm0 (:$ arm64::node-size)) ; ppc:209 caller's vsp
+      ;; W6/KI-026 co-design (was ppc:205 mflr loc-pc): the caller's
+      ;; return pc never rides in temp4 (node-scanned).  Build FRAME-A
+      ;; here, 0221 stp-pair shape; lr still holds the caller's pc and
+      ;; imm1 is call-subprim's scratch, free here.
+      (mov imm1 (:$ arm64::lisp-frame-marker))
+      (stp imm1 imm0 (:@! sp (:$ -32)))
+      (stp fn lr (:@ sp (:$ 16)))
       (call-subprim .SPlexpr-entry)         ; ppc:210 (bla) — DECIDE-10
       (mov lr temp4)                        ; ppc:211 (mtlr loc-pc) — lexpr cleanup continuation
       (mov arg_z vsp)                       ; ppc:212 lexpr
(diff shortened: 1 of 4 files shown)


xrme:
Landed on arm64 as 76ed388c..e3e31a74 (rebased onto arm64), with two follow-up commits unifying the frame-build macro / collapsing the pc_luser_xp case and adding set_xpLR.

Thank you for covering this important case.

PR #615, opened 2026-08-25
arm64: register the fake-stack-frame istruct class

### LINKED ISSUE #610: [arm64] Fault in CCL::CLASS-CELL-TYPEP during a backtrace
This was the second failure reported in #600 and is still present although it no longer happens with the backtrace of `(/ 0 0)`.

Here's an example
```
[palter@minime-debian ~/CCL/ccl-arm64](3)$ ./arm64cl 
Clozure Common Lisp Version 1.13 (v1.13-397-gf0267f7c) LinuxARM6464

For more information about CCL, please see https://ccl.clozure.com.

CCL is free software.
It is distributed under the terms of the Apache License, Version 2.0.
? (#_memset 0 42 0)
> Error: The value 0 is not of the expected type MACPTR.
> While executing: %FF-CALL, in process listener(1).
> Type :POP to abort, :R for a list of available restarts.
> Type :? for other options.
1 > :b
Origin #<error printing 
> Error: Fault during read of memory address #x30000000015BF4
> While executing: CCL::CLASS-CELL-TYPEP, in process listener(1).
> Type :POP to abort, :R for a list of available restarts.
> Type :? for other options.
2 > 
```

And, again, this causes an infinite error loop when using `--batch`
```
[palter@minime-debian ~/CCL/ccl-arm64](3)$ ./arm64cl --batch
Clozure Common Lisp Version 1.13 (v1.13-397-gf0267f7c) LinuxARM
For more information about CCL, please see https://ccl.clozure.com.

CCL is free software.
It is distributed under the terms 6464
? of the Apache License, Version 2.0.
(#_memset 0 42 0)
> Error of type TYPE-ERROR: The value 0 is not of the expected type MACPTR.
> While executing: %FF-CALL, in process listener(1).
 (FFFFBAABC880) : 0 (PRINT-CALL-HISTORY :CONTEXT NIL :PROCESS NIL :ORIGIN NIL :DETAILED-P T :COUNT 1152921504606846975 :START-FRAME-NUMBER 0 :STREAM #<SYNONYM-STREAM to *TERMINAL-IO* #x302000A32BFC> :PRINT-LEVEL 2 :PRINT-LENGTH 5 :PRINT-STRING-LENGTH :DEFAULT :SHOW-INTERNAL-FRAMES NIL :FORMAT :TRADITIONAL) 836
  (&KEY :CONTEXT :PROCESS :ORIGIN :DETAILED-P :COUNT :START-FRAME-NUMBER :STREAM :PRINT-LEVEL :PRINT-LENGTH :PRINT-STRING-LENGTH :SHOW-INTERNAL-FRAMES :FORMAT)
   CCL::CONTEXT: NIL
   PROCESS: NIL
   CCL::ORIGIN: NIL
   CCL::DETAILED-P: T
   COUNT: 1152921504606846975
   CCL::START-FRAME-NUMBER: 0
   STREAM: #<SYNONYM-STREAM to *TERMINAL-IO* #x302000A32BFC>
   CCL::PRINT-LEVEL: 2
   CCL::PRINT-LENGTH: 5
   CCL::PRINT-STRING-LENGTH: :DEFAULT
   CCL::SHOW-INTERNAL-FRAMES: NIL
   FORMAT: :TRADITIONAL

  #:COMPILER-VAR: NIL
  #:COMPILER-VAR: NIL
  #:COMPILER-VAR: NIL
  #:COMPILER-VAR: NIL
  #:COMPILER-VAR: NIL
  #:COMPILER-VAR: NIL
  #:COMPILER-VAR: NIL
  #:COMPILER-VAR: NIL
  #:COMPILER-VAR: NIL
  #:COMPILER-VAR: NIL
  #:COMPILER-VAR: NIL
  #:COMPILER-VAR: NIL
  CCL::FRAME-NUMBER: 0
  *BACKTRACE-PRINT-LEVEL*: 2
  *BACKTRACE-PRINT-LENGTH*: 5
  *BACKTRACE-PRINT-STRING-LENGTH*: :DEFAULT
  *BACKTRACE-FORMAT*: :TRADITIONAL
  *STANDARD-OUTPUT*: #<SYNONYM-STREAM to *TERMINAL-IO* #x302000A32BFC>
  *PRINT-CIRCLE*: NIL

 (FFFFBAABC8A0) : 1 (ABNORMAL-APPLICATION-EXIT) 180
  NIL

  #:G52053: (ERROR)
  CCL::%HANDLERS%: ((ERROR) (PROCES> Error of type CCL::INVALID-MEMORY-ACCESS: Fault during read of memory address #x30000000015BF4
> While executing: CCL::CLASS-CELL-TYPEP, in process listener(1).
S-RESET))

 (FFFFBAABADC0) : 0 (PRINT-CALL-HISTORY :CONTEXT NIL :PROCESS NIL :ORIGIN NIL :DETAILED-P T :COUNT 1152921504606846975 :START-FRAME-NUMBER 0 :STREAM #<SYNONYM-STREAM to *TERMINAL-IO* #x302000A32BFC> :PRINT-LEVEL 2 :PRINT-LENGTH 5 :PRINT-STRING-LENGTH :DEFAULT :SHOW-INTERNAL-FRAMES NIL :FORMAT :TRADITIONAL) 836
  (&KEY :CONTEXT :PROCESS :ORIGIN :DETAILED-P :COUNT :START-FRAME-NUMBER :STREAM :PRINT-LEVEL :PRINT-LENGTH :PRINT-STRING-LENGTH :SHOW-INTERNAL-FRAMES :FORMAT)
   CCL::CONTEXT: NIL
   PROCESS: NIL
   CCL::ORIGIN: NIL
   CCL::DETAILED-P: T
   COUNT: 1152921504606846975
   CCL::START-FRAME-NUMBER: 0
   STREAM: #<SYNONYM-STREAM to *TERMINAL-IO* #x302000A32BFC>
   CCL::PRINT-LEVEL: 2
   CCL::PRINT-LENGTH: 5
   CCL::PRINT-STRING-LENGTH: :DEFAULT
   CCL::SHOW-INTERNAL-FRAMES: NIL
   FORMAT: :TRADITIONAL

  #:COMPILER-VAR: NIL
  #:COMPILER-VAR: NIL
  #:COMPILER-VAR: NIL
  #:COMPILER-VAR: NIL
  #:COMPILER-VAR: NIL
  #:COMPILER-VAR: NIL
  #:COMPILER-VAR: NIL
  #:COMPILER-VAR: NIL
  #:COMPILER-VAR: NIL
  #:COMPILER-VAR: NIL
  #:COMPILER-VAR: NIL
  #:COMPILER-VAR: NIL
  CCL::FRAME-NUMBER: 0
  *BACKTRACE-PRINT-LEVEL*: 2
  *BACKTRACE-PRINT-LENGTH*: 5
  *BACKTRACE-PRINT-STRING-LENGTH*: :DEFAULT
  *BACKTRACE-FORMAT*: :TRADITIONAL
  *STANDARD-OUTPUT*: #<SYNONYM-STREAM to *TERMINAL-IO* #x302000A32BFC>
  *PRINT-CIRCLE*: NIL

 (FFFFBAABADE0) : 1 (ABNORMAL-APPLICATION-EXIT) 180
  NIL

  #:G52053: (ERROR)
  CCL::%HANDLERS%: ((ERR> Error of type CCL::INVALID-MEMORY-ACCESS: Fault during read of memory address #x30000000015BF4
> While executing: CCL::CLASS-CELL-TYPEP, in process listener(1).
OR) (ERROR) (PROCESS-RESET))

...
```

cyclistmass:
Closes #610.

After any error that a trap raises, a backtrace faults:

```
> Error: Fault during read of memory address #x30000000015BF4
> While executing: CCL::CLASS-CELL-TYPEP
```

The address is the same on the reporter's `v1.13-397` build and on mine, because
it is deterministic. The chain, each step observed in a live image:

1. On arm64, as on PPC, the trap-support callbacks pass `%error` a
   `FAKE-STACK-FRAME` istruct as its error-pointer. x86-64 passes a fixnum frame
   pointer, which is why x86-64 never faults here.
2. Backtrace argument recovery is correct. The `%error` frame's symbol map homes
   `ERROR-POINTER` at vsp slot 1, and the recovered value is a valid, live
   fake-stack-frame istruct.
3. `map-entry-value` asks `(typep value 'value-cell)` about every recovered
   value, at `lib/backtrace.lisp:354`. `class-cell-typep` takes the wrapper from
   `(istruct-cell-info (%svref instance 0))`.
4. Nothing on arm64 calls `make-istruct-class` for `CCL::FAKE-STACK-FRAME`, so
   the cell info is `NIL`. `class-cell-typep` then chases a `NIL` wrapper at
   safety 0. The reads are relative to `NIL` in static space, so the faulting
   address is a per-architecture constant.

Two files in the tree call `make-istruct-class`: `compiler/ARM/arm-arch.lisp` and
`level-1/l1-clos-boot.lisp`. ARM32 registers this istruct at
`compiler/ARM/arm-arch.lisp:1446`. arm64 does not.

Neither does PPC, so PPC carries the same latent fault. I have not tested that
and cannot: there is no PPC machine here.

This patch adds the one form to `compiler/ARM64/arm64-arch.lisp`, matching ARM32.
The istruct name is the CCL-package symbol, from `lispequ.lisp`'s
`%cons-fake-stack-frame`.

**Control.** Both halves built and run on linuxarm64 at `d15b911b`, the commit
before this branch tip. The two builds are identical apart from this patch, and
the kernel binary is the same on both sides. The boot image is byte-identical
too, because this is a compiler file and the change lands in the fasl set. The
fasl sets differ.

- RED, this patch withheld:
  `(typep (ccl::%istruct 'ccl::fake-stack-frame 0 0 0 0 0 0 0) 'ccl::value-cell)`
  gives `Fault during read of memory address #x30000000015BF4` in
  `CLASS-CELL-TYPEP`, and no later form in the probe runs.
- GREEN, this patch applied: the same form returns `NIL`, and
  `print-call-history` inside a `handler-bind` for an error that a trap
  raises completes, producing a 6046-character backtrace.

On the GREEN build the ANSI suite reports 21679 tests and 0 failures, and the
CCL-specific suite (`tests/ccl.lsp`) reports 243 tests and 0 failures.

The patch also applies with `git apply --check` to `2bf519ce` itself, which
changes only comments in `level-0/ARM64/arm64-clos.lisp`.

**Not taken here.** `class-cell-typep` has no guard for a `NIL` wrapper, so any
unregistered istruct name reproduces this with
`(typep (%istruct '<name> 0) '<some-class>)`. A guard there would turn a fault
into a type error for every such name. That is wider than this issue needs, and
it belongs with `l1-clos-boot.lisp` rather than with an arm64 fix.

`VERIFIED:` built and run on linuxarm64. The file is under `compiler/ARM64/`, so
no other target is affected. Not built on any other target.

### DIFF
--- a/compiler/ARM64/arm64-arch.lisp
+++ b/compiler/ARM64/arm64-arch.lisp
@@ -1426,3 +1426,11 @@
 ;;; ".SPRPLACD not found" while the variable held it.  Refresh the
 ;;; struct so both lookup paths see the extended vector.
 (setf (arch::target-subprims-table *arm64-target-arch*) *subprims*)
+
+;;; The trap-support callbacks hand %error a FAKE-STACK-FRAME istruct as
+;;; the error-pointer (the PPC model; lispequ.lisp %cons-fake-stack-frame
+;;; conses the CCL-package name).  Give that istruct a class, the way
+;;; arm-arch.lisp does for ARM32's fake-stack-frame.  Without it,
+;;; class-cell-typep chases the istruct's NIL class wrapper and faults,
+;;; and a backtrace after any trap-raised error dies (issue #610).
+(ccl::make-istruct-class 'ccl::fake-stack-frame ccl::*istruct-class*)


xrme:
This is a necessary fix; thanks for finding it.

I included it in my commit 6a52f2ae, which also corrects a misunderstanding about how the control stack is really laid out on arm64.

Issue #12, opened 2017-02-18
Separate 32-bit and 64-bit x86 backends


xrme:
Currently, there is one common x86 backend.

I believe it would be better for them to go their separate ways, and not incidentally, keep out of each other's way.

PR #618, opened 2026-08-27
arm64: four codegen fixes

cyclistmass:
Four independent fixes in the arm64 backend, one file family, each separately
revertible. Split them if you prefer to review them apart.

**1. No FPR scratch in the complex-double-float 128-bit accessors.**
`(+ #C(1d0 2d0) #C(3d0 4d0))` returns `#C(3d0 2d0)`. The rule is
`result.realpart = a.realpart OP a.imagpart` and
`result.imagpart = a.imagpart`. The compiler never reads `b`. Subtraction has
the same shape. `*` is correct.

No other backend declares an FPR temp for this. x86-64 and x86-32 move the
whole 128 bits with one `movdqu`. ARM32 needs a scratch but takes a GPR, `lr`,
which it declares through `:sets-lr`. arm64 composes it from
two D-form `LDUR`s plus an `INS` through a `(dtemp :double-float)`, and that
scratch is not safe. `with-fp-target` and `available-fp-temp` only choose a
register. `compiler/backend.lisp:492`, above `with-imm-target`, which
`with-fp-target` copies: *"Choose an immediate register (for targeting), but
don't 'reserve' it."* Neither macro assigns `*available-backend-fp-temps*`, and
no FPR-targeting path in `arm642.lisp` calls `use-fp-reg`. So the scratch
avoids that vinsn's own operands and nothing else, in particular not the
caller's other live operand.

**2. A fixnum literal in a numeric comparison becomes a cmp-immediate.**
`arm642-inline-numcmp` takes its constant path only for `(= x 0)`. So
`(< i 10)` computes 10 into a register, tag-checks both operands, then does a
register-register compare. `x862-inline-numcmp` admits any constant that fits
its immediate, for every predicate, and reverses the condition when the
constant sits on the left.

This does the same, and both pieces already exist.
`compare-signed-s16const` encodes the whole signed range but has one caller
passing it 0. `arm642-swap-compare-cond-bit` sits at `arm642.lisp:9787` with no
caller anywhere in the tree, and it is the operand-order reversal this needs.
The patch gates the constant to `|boxed| < 4096`, so `|c| < 512` at fixnumshift
3, which covers a literal loop bound. A wider gate is a separate change.

**3. A negative constant array index must not stay `index-known-fixnum`.**
`acode-fixnum-form-p` returns the VALUE, so a negative literal makes the flag
non-nil and the compiler treats the index as a trusted displacement. At
`safety 0` both arms then go wrong at once. `unscaled-idx` stays `NIL`,
`check-misc-bound` never runs, and `vref1` gets the negative constant as its
displacement. The access lands at a fixed offset BEFORE the vector data. At
index -1 that is the uvector header.

With the guard deleted in a live image, read back through the resident
disassembler: the specialized element types raise a template constraint error,
and `(svref v -1)` silently emits `(ldur arg_z (:@ arg_z (:$ -12)))` and
returns the header as an object.

**4. `add-immediate` must declare the range its body can encode.**
It declares its constant operand `:s32const`, and `compiler/vreg.lisp` maps
that to `(signed-byte 32)`. The body reads two 12-bit fields, so it reaches 24
bits of magnitude and no more. At `2^24` both fields are zero and the body
emits `add dest,src,#0,lsl #12`, which adds ZERO. The class check is the only
instrument in that path, and the declaration tells it to stay silent.

The defect is LATENT, not live: all three emit sites gate the constant first,
at 12, 24 and 24 bits. The patch also touches one line of `compiler/vreg.lisp`,
the portable half of the same declaration.

---

**Verified at this base.** Built and run on linuxarm64 at `ec578745` with all
ten patches applied: ANSI 21679 tests, 0 failures. `tests/ccl.lsp` 243 tests, 0
failures. Image `281a49e5`, kernel `67bb66b4`.

### DIFF
--- a/compiler/ARM64/arm64-vinsns.lisp
+++ b/compiler/ARM64/arm64-vinsns.lisp
@@ -1726,9 +1726,15 @@
 ;;; s61-truncation class).  Lane order low-then-high; the first emitted
 ;;; insn reads src, any second insn reads dest -- dest=src safe.
 ;;; const=0 degenerates to (add dest src #0) -- a copy, correct.
+;;; The class is :s24const because AArch64 ADD (immediate) holds imm12 with an
+;;; optional LSL #12, so the two lanes below reach 24 bits of magnitude and no
+;;; more.  A :s32const declaration admits a constant the body then drops: at
+;;; 2^24 both ldb fields are 0 and the vinsn emits `add dest,src,#0,lsl #12',
+;;; which adds ZERO with no diagnostic.  (signed-byte 24) is the conservative
+;;; fit, and every emit site already gates at (signed-byte 24) or tighter.
 (define-arm64-vinsn add-immediate (((dest :imm))
                                    ((src :imm)
-                                    (const :s32const)))
+                                    (const :s24const)))
   ((:pred >= const 0)
    ((:not (:pred = 0 (:apply ldb (byte 12 0) const)))
     (add dest src (:$ (:apply ldb (byte 12 0) const)))
--- a/compiler/ARM64/arm64-vinsns.lisp
+++ b/compiler/ARM64/arm64-vinsns.lisp
@@ -2189,25 +2189,19 @@
 ;;; ============ misc-ref-c-complex-double-float ============
 ;;; PPC64 ppc64-vinsns.lisp:271: two lfd loads (realpart @
 ;;; complex-double-float.realpart + 16k, imagpart +8) into an FPR pair.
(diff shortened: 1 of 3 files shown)


xrme:
I look the first three commits as 429428504, 6c9f11fa7e, and 3815d1f39a0.  A related commit is 6c3c2e7b2, where I narrowed some vreg specs to be closer to actual maximums.

I declined the fourth commit here because I look at vreg specs as things meant to model machine widths; they're not a general typep-like mechanism.  Users of the vinsns will have to maintain tighter values ranges on their own.

Thank you for the changes.

PR #630, opened 2026-09-07
Fix allocate_list not considering that GC may relocate initial-element on arm64 and ppc

yaso9:
`allocate_list` did not consider that if the GC runs during execution,
the initial-element may be moved. Since it held onto the pre-relocation
pointer, this could cause corruption. On qemu-system-aarch64 running
Debian Trixie, this can be triggered by the following program:

```common-lisp
  (defparameter *n* 110000)
  (defparameter *a* (list 1 2 3))
  (defparameter *b* (make-list *n* :initial-element *a*))
  (defparameter *c* (make-list *n* :initial-element *a*))
```

Executing:

```common-lisp
  (format t "Equal car? ~A~%" (equal (car *b*) (car *c*)))
```

Yields:

```
  Equal car? NIl
```

This appears to affect the ppc backend as well. The x86 backend appears
to have already made this correction. See the following in
x86-exceptions.c:

```c
  /*
  * Note that the GC may relocate the initial element stored in
  * arg_y, so it's not safe to save arg_y in a local variable.
  */
  deref(current, 1) = xpGPR(xp, Iarg_y);
```

I have tested the arm64 change and it fixes the above reproduction. I do not have a PPC machine to test the PPC change. I have separated the fixes into separate commits to be able to drop the PPC fix if requested.

### DIFF
--- a/lisp-kernel/arm64-exceptions.c
+++ b/lisp-kernel/arm64-exceptions.c
@@ -536,8 +536,7 @@ allocate_list(ExceptionInformation *xp, TCR *tcr)
     bytes_needed = (nconses << dnode_shift);
   LispObj
     prev = lisp_nil,
-    current,
-    initial = xpGPR(xp,arg_y);
+    current;
 
   if (nconses == 0) {
     /* Silly case */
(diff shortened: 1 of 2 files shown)


xrme:
Nice catch; thank you.

Issue #597, opened 2026-08-09
recursive_lock_trylock leaks a recursive count; :futex is disabled for every target


cyclistmass:
Two things in the futexless lock code, found at 18873bd1 while chasing a
customer report on linuxarm64. Neither is urgent. The first is dead code today.

**1. `recursive_lock_trylock` leaks a recursive count.**

`thread_manager.c:399-421`, and the `USE_FUTEX` arm at `:425` has the same shape:

```c
  if (owner == tcr) {
    m->count++;
    if (was_free) {
      *was_free = 0;
      RELEASE_SPINLOCK(m->spinlock);
      return 0;
    }
  }
  if (store_conditional((natural*)&(m->avail), 0, 1) == 0) {
```

When `owner == tcr` and `was_free` is NULL, it increments `m->count` and falls
through. The `store_conditional` then fails, because we already hold the lock,
so the function returns EBUSY with the count already raised. A leaked count
makes the lock permanently owned: every other thread blocks, and the owner sees
a recursive acquire that never drains.

It looks like a missing `else`.

This is not reachable today. Nothing in the tree calls `recursive_lock_trylock`,
and the Lisp side names it only in the kernel-import enum in each arch file. So
it is latent, and I have not observed it.

**2. `:futex` is off for every target.**

`level-0/l0-misc.lisp:20`:

```lisp
#+(and linux-target no (or x86-target arm-target))
(eval-when (:compile-toplevel :execute)
  (pushnew :futex *features*))
```

`no` is not a feature, so the form never fires. Every Linux build runs the
`#-futex` protocol. There is a second reason as well: the guard names
`arm-target`, not `arm64-target`.

I mention it because it changes who the `#-futex` code belongs to. I had assumed
stock Linux CCL never ran it, and that is wrong. It is the live protocol on
linuxarm64 and, if I read the gate correctly, on linux x86-64 too.

If the `no` is deliberate, say so and I will stop reading it as a leftover.

Mauro DiBenedetto <maurodibenedetto@gmail.com>

cyclistmass:
Correcting myself on the fix, not on the defect.

I wrote that this looks like a missing `else`. That is wrong, and it would send
you to a change that does not fix it. With `} else if (store_conditional...)`,
the `owner == tcr` and `was_free == NULL` path still falls through to
`return EBUSY` with the count already raised.

The defect is the early `return 0`. It is gated on `was_free != NULL`, when only
the `*was_free = 0` store should be:

```c
[...]

PR #634, opened 2026-09-09
A thread suspended inside a lock guard deadlocks the thread that suspended it

cyclistmass:
A thread can take the suspend signal while it holds one of the spin words that
guard a lock structure. It parks there. The thread that stopped the world then
needs that same word to release the exception lock. The holder cannot run until
somebody resumes it, and nobody does.

One core spins at 100 percent until somebody kills the process. No crash, no
error, no debugger entry. We have four field occurrences, and the longest wedged
for six days.

## The cycle

`%suspend-other-threads` traps to `XUUO_SUSPEND_ALL`. That handler is the only
normal-return caller of `suspend_other_threads` that returns with the world
still stopped.

Meanwhile another thread takes an allocation trap. There
`prepare_to_wait_for_exception_lock` calls `ALLOW_EXCEPTIONS`, which restores
the interrupted signal mask. That thread then takes the recursive lock's guard
word in `lock_recursive_lock`. The signal is deliverable there, and nothing
marks the spin sections. `suspend_resume_handler` defers only at
`TCR_INTERRUPT_LEVEL <= -(2 << fixnumshift)`. So the thread parks holding the
word. The suspender reaches `unlock_exception_lock_in_handler` and spins on it
forever.

Two constraints on a fix. `ALLOW_EXCEPTIONS` before the wait is deliberate, and
the comment there says why. We cannot mask the signal across the wait.

The word also does not name its owner. `get_spin_lock` swaps the caller's TCR in
on every attempt. A self-ownership test is therefore unsound.

## The fix

`with-deferred-gc` binds `*interrupt-level*` to -2. The documentation says it runs
"without responding to the signal used to suspend threads", and
`suspend_resume_handler` honors that threshold. Nobody ever applied the
mechanism to the runtime's own lock protocol, which runs at -1.

Each patch raises the level across guard acquisition and the work held under it.
It then releases, restores the previous level, and re-raises a pending suspend.
Restoring before polling is race-free. A signal before the restore sets
`PENDING_SUSPEND` for the poll to find, and one after parks normally.

- `thread_manager.c` (+83/-3), all eight non-futex routines. The helpers skip a
  TCR with no `tlb_pointer`. The initial thread takes one of these locks in
  `new_tcr`, before its thread-local vector exists. An earlier version crashed
  the boot there.
- `level-0/l0-misc.lisp` (+106/-28), the same protocol on the Lisp side. Nine
  acquisitions in seven routines, all previously under `without-interrupts`,
  which is -1 and does not defer suspension. Two fixes ride along.
  `%unlock-rwlock-ptr` now posts its semaphores after releasing the word.
  `%promote-rwlock` no longer returns from its owner arm still holding it.
- `arm64-spentry.s` (+32/-36) and `x86-spentry64.s` (+15/-15) fix
  `unbind_interrupt_level`. It tests `PENDING_SUSPEND` before restoring the
  level, then acts on the value it already loaded. A suspend arriving in that
  window is lost. PPC has the same bug and no patch, since nobody can build or
  test it.

The first commit is the `recursive_lock_trylock` count leak from #597. The
already-owned path raises the count and then falls through to EBUSY. It is here
because the deferral change touches that function.

That commit stands alone as a fix, and the deferral commit does not apply
without it. The leak fix moves `RELEASE_SPINLOCK` and `return 0` out of the
`if (was_free)` block, and the deferral commit adds `allow_thread_suspension`
against that corrected shape. Take the two together or take neither.

## Testing

**A correction to an earlier version of this section.** The driver behind the
per-reproducer counts printed its success line from the environment variable it
was handed rather than from the log, so "3000 of 3000" in both places and the
20,000 soak stated the value passed in. Those runs may well have completed. The
numbers do not establish that they did, and they are gone. The red results
below parse the log for a heartbeat stall, and the gdb captures are direct
observations, so both stand.

The strongest result needs no widener at all. Stock CCL 1.13 LinuxX8664, one
2-vCPU x86-64 host, no instrumentation of any kind:

| build | result |
|---|---|
| stock, unpatched | RED 3 of 3, wedged at 39,500 / 164,500 / 200,250 cycles |
| with the two deferral patches | clean 3 of 3 to the 299,750 cap, and 625,500 in a longer run |

Every wedge shows one thread runnable and spinning in
`unlock_exception_lock_in_handler` -> `unlock_recursive_lock` -> `get_spin_lock`,
with 25 sleeping in `suspend_resume_handler` and `sem_wait_forever`.

The wedge point is stochastic over an order of magnitude. Ten runs wedge at
18,500 / 28,500 / 39,500 / 51,750 / 70,500 / 88,250 / 103,250 / 137,000 /
164,500 / 200,250 cycles, and one unpatched run survived to 199,750 before I
stopped it. A short clean run proves nothing here.

Three reproducers, each watched red and then green, all on Linux x86-64.

The first covers the C side. It runs N threads allocating hard while one loops a
small `with-other-threads-suspended` body. The window is about ten instructions,
so a throwaway spin inside `LOCK_SPINLOCK` widens it, gated on an environment
variable. Widened and unfixed, it wedges before the first heartbeat. `thread
apply all bt` then shows a worker parked in `suspend_resume_handler` on top of
`get_spin_lock`, which is the field signature. With the C patch it stops
wedging.

The second covers the Lisp side and uses read locks as the vehicle. A reader
that owns the lock never blocks another reader, so a suspended reader-owner
cannot stall the world-stopper by design. That leaves the guard word as the only
shared serialization point. With the Lisp patch withheld, the run stalls
permanently and never reaches a heartbeat. The capture shows the readers in the
`%get-spin-lock` spin loop, and no thread waiting on a lock semaphore. With the
patch restored, the stall does not return.

Nothing else differs between the two runs.

We watched every suite result below twice, and kernel identity moved at every
step.

- x86-64: ANSI 21679 tests 0 failed, `ccl.lsp` 243 tests 0 failed, on eleven
  consecutive runs.
- arm64, full build from a clean reset: ANSI 21679 tests 0 failed, `ccl.lsp` 243
  tests 0 failed.

The third covers `unbind_interrupt_level`, and its observable is different. A
missed suspend there is not permanent. The victim self-heals at its next unbind,
so the cost is a world-stop that waits for it. The reproducer therefore times
every world-stop rather than watching for a wedge. A sled widens the window at
the pre-fix instruction order. A matching sled sits at the post-fix point. Both
runs therefore carry the same added delay, and they differ only in where the
window falls.

Pre-fix, 2000 iterations: 2 missed suspends, longest world-stop 157 ms. Post-fix,
same workload and the same added delay: 0 missed, longest 19 ms. Two events are
not a rate, and I am not offering one. What the run shows is the latency
separation, and it agrees with the mechanism. Before the fix, the window spans
the flags read to the restore, so a signal landing in it defers against a stale
read. After the fix, it spans the restore to the poll, where a signal parks at
once.

Nothing here reproduces on arm64. For the C and Lisp patches that is a thin gap,
since the code is shared and only the host differs. For `arm64-spentry.s` it is a
real one. That file is its own assembly. The arm64 half of the unbind fix rests
on the ordering argument, on matching the x86-64 change instruction for
instruction, and on a full arm64 build with both suites.

The `recursive_lock_trylock` commit has no test of its own either. It is the
#597 fix, unchanged, and it rests on reading and on the suites.

All three reproducers and their wideners are test-only, so I keep them out of
the series. They now live in Clozure/ccl-tests#10 under `extended-tests/`.

## Related

#597 also reports that `:futex` never turns on. The feature guard reads
`#+(and linux-target no ...)`, and there is no feature named `no`. That is why
the non-futex protocol is live code on every Linux target. Enabling futex
removes the guard word and this whole class of defect, and it is a separate
decision.

Gary Palter asked us to submit this rather than hold it.

## Where these tests live

This reproducer had no home. It is not an ANSI test. It needs several threads, a
widened window and a gdb capture. We have a number of tests in that shape from
the arm64 port. Each came from a real defect that a full ANSI run passed, and
each passes on stock x86-64. They cover declared element types at speed 3 safety 0,
FFI and callbacks, threads and interrupts, GC and purify, backtrace, and fasl
round-trips.

Answered, and done. Clozure/ccl-tests#10 adds `extended-tests/` with these
reproducers in it.

### DIFF
--- a/level-0/l0-misc.lisp
+++ b/level-0/l0-misc.lisp
@@ -578,6 +578,33 @@
   (defun %release-spin-lock (p)
     (setf (%get-natural p 0) 0)))
 
+;;; Take the spin word at P with thread suspension deferred, so a
+;;; world-stop cannot park this thread while it holds the word: a
+;;; suspended holder never releases it, and a thread that then needs
+;;; the word spins forever.  An *INTERRUPT-LEVEL* of -2 defers
+;;; suspension, exactly as WITH-DEFERRED-GC does.  The deferral covers
+;;; only each batch of atomic attempts; failed batches deliver any
+;;; pending suspend and yield at the enclosing WITHOUT-INTERRUPTS
+;;; level, because YIELD dispatches and makes a foreign call.  Returns
+;;; with *INTERRUPT-LEVEL* -2 and the word held; the caller restores
+;;; the level immediately after releasing the word and then calls
+;;; %CHECK-DEFERRED-GC, before any wait, error, semaphore post or
+;;; nested lock operation.
+#-futex
+(defun %get-spin-lock-deferring-suspension (p)
+  (let* ((self (%current-tcr))
+         (n *spin-lock-tries*))
+    (declare (fixnum n))
+    (loop
+      (setq *interrupt-level* -2)
+      (dotimes (i n)
+        (when (eql 0 (%ptr-store-fixnum-conditional p 0 self))
+          (return-from %get-spin-lock-deferring-suspension t)))
+      (setq *interrupt-level* -1)
+      (%check-deferred-gc)
+      (%atomic-incf-node 1 '*spin-lock-timeouts* target::symbol.vcell)
+      (yield))))
+
 (eval-when (:compile-toplevel :execute)
   (declaim (inline note-lock-wait note-lock-held note-lock-released)))
 
@@ -604,15 +631,19 @@
          (when flag
            (setf (lock-acquisition.status flag) t))
          (return t))
-       (%get-spin-lock spin)
+       (%get-spin-lock-deferring-suspension spin)
        (when (eql 1 (incf (%get-natural ptr target::lockptr.avail)))
          (setf (%get-ptr ptr target::lockptr.owner) p
                (%get-natural ptr target::lockptr.count) 1)
          (%release-spin-lock spin)
+         (setq *interrupt-level* -1)
+         (%check-deferred-gc)
          (if flag
            (setf (lock-acquisition.status flag) t))
          (return t))
-       (%release-spin-lock spin))
+       (%release-spin-lock spin)
+       (setq *interrupt-level* -1)
+       (%check-deferred-gc))
       (%process-wait-on-semaphore-ptr signal 1 0 (recursive-lock-whostate lock)))))
 
 #+futex
@@ -718,12 +749,14 @@
               t)
              (t
               (let* ((win nil))
-                (%get-spin-lock spin)
+                (%get-spin-lock-deferring-suspension spin)
                 (when (setq win (eql 1 (incf (%get-natural ptr target::lockptr.avail))))
                   (setf (%get-ptr ptr target::lockptr.owner) p
                         (%get-natural ptr target::lockptr.count) 1)
                   (if flag (setf (lock-acquisition.status flag) t)))
                 (%release-spin-lock spin)
+                (setq *interrupt-level* -1)
+                (%check-deferred-gc)
                 win)))))))
 
 
@@ -762,7 +795,7 @@
     (without-interrupts
      (when (eql 0 (decf (the fixnum
                           (%get-natural ptr target::lockptr.count))))
-       (%get-spin-lock spin)
+       (%get-spin-lock-deferring-suspension spin)
        (setf (%get-ptr ptr target::lockptr.owner) (%null-ptr))
        (let* ((pending (+ (the fixnum
                             (1- (the fixnum (%get-fixnum ptr target::lockptr.avail))))
@@ -771,6 +804,8 @@
          (setf (%get-natural ptr target::lockptr.avail) 0
                (%get-natural ptr target::lockptr.waiting) 0)
          (%release-spin-lock spin)
+         (setq *interrupt-level* -1)
+         (%check-deferred-gc)
          (dotimes (i pending)
            (%signal-semaphore-ptr signal)))))
     nil))
@@ -878,11 +913,13 @@
            (tcr (%current-tcr)))
       (declare (fixnum tcr))
       (without-interrupts
-       (%get-spin-lock ptr)               ;(%get-spin-lock (%inc-ptr ptr target::rwlock.spin))
+       (%get-spin-lock-deferring-suspension ptr)
        (if (eq (%get-object ptr target::rwlock.writer) tcr)
          (progn
            (incf (%get-signed-natural ptr target::rwlock.state))
            (%release-spin-lock ptr)
+           (setq *interrupt-level* -1)
+           (%check-deferred-gc)
            (if flag
              (setf (lock-acquisition.status flag) t))
            t)
@@ -891,15 +928,19 @@
                ;; That wasn't so bad, was it ?  We have the spinlock now.
                (setf (%get-signed-natural ptr target::rwlock.state) 1)
                (%release-spin-lock ptr)
+               (setq *interrupt-level* -1)
+               (%check-deferred-gc)
                (%set-object ptr target::rwlock.writer tcr)
                (if flag
                  (setf (lock-acquisition.status flag) t))
                t)
            (incf (%get-natural ptr target::rwlock.blocked-writers))
            (%release-spin-lock ptr)
+           (setq *interrupt-level* -1)
+           (%check-deferred-gc)
            (let* ((*interrupt-level* level))
                   (%process-wait-on-semaphore-ptr write-signal 1 0 (rwlock-write-whostate lock)))
-           (%get-spin-lock ptr)))))))
+           (%get-spin-lock-deferring-suspension ptr)))))))
 #+futex
 (defun %write-lock-rwlock-ptr (ptr lock &optional flag)
   (with-macptrs ((write-signal (%INC-ptr ptr target::rwlock.writer-signal)) )
@@ -952,10 +993,12 @@
            (tcr (%current-tcr)))
       (declare (fixnum tcr))
       (without-interrupts
-       (%get-spin-lock ptr)             ;(%get-spin-lock (%inc-ptr ptr target::rwlock.spin))
+       (%get-spin-lock-deferring-suspension ptr)
        (if (eq (%get-object ptr target::rwlock.writer) tcr)
          (progn
            (%release-spin-lock ptr)
+           (setq *interrupt-level* -1)
+           (%check-deferred-gc)
            (error 'deadlock :lock lock))
          (do* ((state
                 (%get-signed-natural ptr target::rwlock.state)
@@ -965,15 +1008,19 @@
                (setf (%get-signed-natural ptr target::rwlock.state)
                      (the fixnum (1- state)))
                (%release-spin-lock ptr)
+               (setq *interrupt-level* -1)
+               (%check-deferred-gc)
                (if flag
                  (setf (lock-acquisition.status flag) t))
                t)
            (declare (fixnum state))
            (incf (%get-natural ptr target::rwlock.blocked-readers))
            (%release-spin-lock ptr)
+           (setq *interrupt-level* -1)
+           (%check-deferred-gc)
            (let* ((*interrupt-level* level))
              (%process-wait-on-semaphore-ptr read-signal 1 0 (rwlock-read-whostate lock)))
-           (%get-spin-lock ptr)))))))
+           (%get-spin-lock-deferring-suspension ptr)))))))
 
 #+futex
 (defun %read-lock-rwlock-ptr (ptr lock &optional flag) 
@@ -1023,19 +1070,26 @@
   (with-macptrs ((reader-signal (%get-ptr ptr target::rwlock.reader-signal))
                  (writer-signal (%get-ptr ptr target::rwlock.writer-signal)))
     (without-interrupts
-     (%get-spin-lock ptr)
+     (%get-spin-lock-deferring-suspension ptr)
      (let* ((state (%get-signed-natural ptr target::rwlock.state))
             (tcr (%current-tcr)))
        (declare (fixnum state tcr))
        (cond ((> state 0)
               (unless (eql tcr (%get-object ptr target::rwlock.writer))
                 (%release-spin-lock ptr)
+                (setq *interrupt-level* -1)
+                (%check-deferred-gc)
                 (error 'not-lock-owner :lock lock))
               (decf state))
              ((< state 0) (incf state))
              (t (%release-spin-lock ptr)
+                (setq *interrupt-level* -1)
+                (%check-deferred-gc)
                 (error 'not-locked :lock lock)))
        (setf (%get-signed-natural ptr target::rwlock.state) state)
+       (let* ((nwriters 0)
+              (nreaders 0))
+         (declare (fixnum nwriters nreaders))
        (when (zerop state)
          ;; We want any thread waiting for a lock semaphore to
          ;; be able to wait interruptibly.  When a thread waits,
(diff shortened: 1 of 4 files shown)


cyclistmass:
Update. Three of the gaps I listed above are now closed. Parts of the
description are out of date, and I would rather say so than leave them standing.

All five commits now have a watched failing test.

**`recursive_lock_trylock`.** The description says this one has no test. It has
one now. Getting there corrected a claim I nearly made to you.

There is no caller of the function anywhere in the tree, so the leak is latent
rather than live. I first read that as "no runtime test is possible". That was
wrong. The kernel-import vector is callable, and level-0 already uses the idiom,
so a test can reach the C function even with no production caller. A search for
callers cannot settle reachability when a dispatch table sits in between.

The test asserts nothing about the hidden count. It takes the lock, calls the
function on the already-owned path with a null `was_free`, then releases the
number of times the contract says to. A second process then tries to acquire.
Unfixed, the raw call returns EBUSY having raised the count, so one release is
left owing and the second process cannot take the lock. Fixed, it does. The
failure is visible from outside the thread under test.

Both patches have to be withheld together for the red, since the deferral change
rewrites the same arm.

**arm64 now reproduces.** The description says it does not. Both arm64 patches
have their own red-then-green on hardware:

- `arm64-spentry.s`: 6000 iterations, 2 missed suspends at 320 ms, against 0
  missed at 32 ms with the fix.
- `level-0/l0-misc.lisp`: the widener cannot reach `%release-spin-lock` there,
  since it is LAP. Widening at runtime instead, against the unmodified kernel
  and image, gives a permanent wedge with zero heartbeats. With the fix, 3000 of
  3000 world-stops. The captures show the same signature as x86-64, including no
  thread waiting on a lock semaphore, which rules out the by-design owner stall.

**One note on the trylock test, stated plainly.** It fails on a stock build,
because a stock build still has this defect. That is the expected state until
the fix merges. I mention it so the result is not mistaken for the test
measuring something local to us.

**One thing I found while reading that is not in this branch.** The inline
`unbind-interrupt-level` vinsn does not check the pending-suspend flag at all,
on both arm64 and x86-64. Not the wrong order. Absent. Both already restore
before they test, so it is a different and larger gap than the one these patches
fix, and the subprimitives do check it. It sits behind an inline declaration
that is off by default, so nothing runs it today. I have not written a fix or a
test for it. Happy to open a separate issue if that is useful.

xrme:
@cyclistmass I think it's a good idea to have a place for tests that are more complicated than the simple regression tests that we have in https://github.com/Clozure/ccl-tests.

What about a new top-level directory in the ccl-tests repository, called extended-tests or other-tests or some suitable name?  I'd like to be able to see and possibly run the reproducers, and they might be useful for testing all the current ports (x86, arm64, 32-bit arm).

Would you care to make PR for that?

cyclistmass:
**Two corrections to the description above, and one answer.**

### 1. Three of the green numbers in Testing are not measurements

The driver behind them printed its success line as `completed $ITERS world-stops`, taken from the environment variable it was handed and not from the log. A run that stopped after 200 iterations printed the same line as a run that finished. So "3000 of 3000" in both places, and the "unwidened soak 20,000 of 20,000", state the value I passed in. Those runs may well have completed. The numbers do not establish that they did, and I would rather retract them than leave them standing.

The red half of that section holds. RED comes from parsing the log for a heartbeat stall, and the gdb captures are direct observations.

### 2. What replaces them needs no widener at all

Stock CCL 1.13 LinuxX8664, one 2-vCPU x86-64 host, no instrumentation of any kind, stock binary:

| build | result |
|---|---|
| stock, unpatched | RED 3 of 3, wedged at 39,500 / 164,500 / 200,250 cycles |
| with the two deferral patches | clean 3 of 3 to the 299,750 cap, and 625,500 in a longer run |

Every wedge shows the same shape. One thread runnable, spinning in `unlock_exception_lock_in_handler` -> `unlock_recursive_lock` -> `get_spin_lock`, and 25 sleeping in `suspend_resume_handler` and `sem_wait_forever`.

The wedge point is stochastic over an order of magnitude. Ten runs: 18,500 / 28,500 / 39,500 / 51,750 / 70,500 / 88,250 / 103,250 / 137,000 / 164,500 / 200,250 cycles. One unpatched run survived to 199,750 cycles before I stopped it. A short clean run therefore proves nothing here, which is why the driver now refuses to print GREEN below 400,000 cycles and prints INCONCLUSIVE instead.

### 3. The two thread_manager commits are not separable

`defer thread suspension while a guard spin lock is held` does not apply without `recursive_lock_trylock leaks a recursive count on EBUSY`. The leak fix moves `RELEASE_SPINLOCK` and `return 0` out of the `if (was_free)` block, and the deferral commit adds `allow_thread_suspension` against that corrected shape, so the deferral patch fails on context if the leak fix is absent. I hit this while building the unpatched baseline for the table above. The description called the leak fix standalone, which is true of the fix and misleading about the pair.

### The answer

The reproducers now live in Clozure/ccl-tests#10, under the `extended-tests/` name you suggested. They run from a checkout with no setup:

```sh
CCL=/path/to/lx86cl64 bash extended-tests/threads/run-suspend-spinlock.sh
```

xrme:
I'm working on this.  I've been investigating the unbind_interrupt_level changes on arm, and 32-bit x86 in addition to arm64 and x86-64.

xrme:
The unbind_interrupt_level changes are in 1606a83d6cc (arm64, x86-64, x86-32) and 53a509a6cf7 (32-bit ARM).

xrme:
One issue I'd like to raise here is that code like
```
  (ccl::with-other-threads-suspended
    (ccl::read-lock-rwlock *rw*)
    (ccl::unlock-rwlock *rw*))
```
is not reasonable.  Manipulating locks inside with-other-threads-suspended is not something that can be done: acquiring a lock in this situation is either a no-op or a deadlock.  The whole idea of with-other-threads-suspended is fragile: anything running under it must not take a lock, and shouldn't allocate.

cyclistmass:
You are right, and the reproducer you quoted is wrong. I withdraw it.

`suspend-spinlock-rwlock.lisp` takes a read lock inside
`with-other-threads-suspended`. By your rule that file tests nothing: it is the
no-op-or-deadlock case, by construction. I will drop it from the pull request
rather than try to repair it.

It also carried an argument, and the argument falls with it. I had used that
file to claim a reader never blocks a reader, so a wedge there could not be the
by-design hazard of a suspended thread holding the lock. That conclusion rested
on a test which breaks the rule, so I withdraw it too. Whether the by-design
case is excluded is open again, and I no longer claim it is.

What I do not think the objection reaches is the main reproducer. Its window is
empty:

```lisp
(ccl::%suspend-other-threads)
(ccl::%resume-other-threads)
```

No lock, no allocation, and nothing printed until after the resume. So the lock
that deadlocks is not one the test takes. It is CCL's own, on the exception
return path: `unlock_exception_lock_in_handler` calls `UNLOCK`, which is
`unlock_recursive_lock`, which takes the lock's spin word through
`LOCK_SPINLOCK`.

That is the question I would put back to you. Does the rule bind the runtime's
own release path as well? If nothing running under a world stop may take a
lock, then releasing the exception lock does the forbidden thing on CCL's
behalf, while the thread that would answer is suspended. If instead that path
is meant to be safe because the lock is the runtime's own, then I report that
it is not — on this host, at this rate.

I would rather have your answer to that than defend the test I got wrong.

xrme:
I've been studying this, and your diagnosis holds up.

As mentioned before, the unbind_interrupt_level changes are in https://github.com/Clozure/ccl/commit/1606a83d6ccb992929e364d5295f96fb1e585799 (arm64, x86-64, x86-32) and https://github.com/Clozure/ccl/commit/53a509a6cf7a9e887794d68becd743c11f07f93b (32-bit ARM).

The ARM change is in a separate commit because it fixes a bad bug (imm1 contained junk) that your restructuring found and fixed.

Your suggestion was to add code to defer suspensions in the generic recursive-lock primitives.  I changed that into a pair of suspend-deferring lock/unlock functions that are only used where they are specifically needed.
 * 04f1e0ac adds `lock_recursive_lock_deferring_suspension` and its unlock counterpart
 * 088e706e converts the `EXCEPTION_LOCK` call sites to `LOCK_EXCEPTION_LOCK` / `UNLOCK_EXCEPTION_LOCK`.

The rationale: the lisp kernel uses two recursive locks: `EXCEPTION_LOCK` and `TCR_AREA_LOCK`.  `EXCEPTION_LOCK` is rather special.  Consider the handler for `XUUO_SUSPEND_ALL`: it returns with other threads suspended, so `unlock_exception_lock_in_handler` grabs the spin lock while other threads are frozen (so the frozen threads had better not hold that spin lock).

`TCR_AREA_LOCK`, on the other hand, is taken around the suspended window.  Nothing tries to take it while the world is stopped.  (Indeed, if we deferred suspension inside locking it, it might be possible for the suspending thread to block while holding it, and there would be no thread around to rescue it.)

Therefore, it makes sense for suspend-awareness to be a property of a particular lock.

I left a separate comment on Clozure/ccl-tests#10 describing how these changes affect your run-suspend-spinlock.sh test (spoiler: they fix it).

The Lisp part of your PR is still unaddressed.  `*kernel-exception-lock*` and `*kernel-tcr-area-lock*` are references to the same memory as the corresponding C structures.  Lisp can still reach these locks (`static-cons`, notably).  So, again, rather than changing the Lisp lock primitives, I think we need to make the suspension-awareness a property of the lock so that ordinary `make-lock` locks stick to the plain path.

I think I'll leave the PR open for now just so that I don't forget about the Lisp part of the job, but it's also true that the Lisp-side fix will end up looking different from your proposed change.

cyclistmass:
Thank you. Making suspend-awareness a property of the lock is the better shape,
and the asymmetry you name is one we missed: `TCR_AREA_LOCK` is held around the
suspended window, so deferring inside it could park the suspending thread with
nothing left to rescue it. The same change that fixes one lock breaks the other.

We measured your fix rather than read it. Two cells on one 2-vCPU Linux x86-64
host, one variable between them: the four commits. Same widener, same image,
same iteration and worker counts. The tree carries the test widener and none of
our lock patches.

| pin | verdict |
|---|---|
| `d226304f`, before your commits | RED, wedged, no heartbeat |
| `088e706e` | GREEN, 3000 world-stops completed |

The red capture matters more than the verdict. It shows a consing worker parked
in `suspend_resume_handler` on top of `prepare_to_wait_for_exception_lock`, with
the initial thread in `sem_wait_forever`. That is the signature of the field
occurrences behind this report, and those were x86-64. So your
`EXCEPTION_LOCK` conversion closes the route we actually hit, and it closes it
for us with no Lisp change at all.

The Lisp half still reproduces at `088e706e`. The read-lock reproducer wedges
with no heartbeat, all four readers and the initial thread parked in
`suspend_resume_handler`. It needs no widener to do it. One caveat on that cell:
we have the red today and not a paired green, because the green needs the
`l0-misc` change you are replacing.

We are glad to write the Lisp half in the shape you describe, with the property
on the lock so `make-lock` locks keep the plain path. Say the word and we will,
or leave it with you if you would rather write it yourself.

Two changes in our `l0-misc` diff are not deferral changes, and we flag them so
they do not get lost either way. `%unlock-rwlock-ptr` posts its semaphores while
it still holds the spin word, and those posts are foreign calls.
`%promote-rwlock` returns from its owner arm still holding the word. Both are
defects on their own terms, whatever the suspension mechanism ends up being.

xrme:
I fixed the Lisp side in 2c382468066e.

I added %lock-recursive-lock-ptr-deferring-suspension and its corresponding unlock counterpart, mirroring the lock_recursive_lock_deferring_suspension code from the lisp kernel.  I changed with-exception-lock to expand into those, and updated all sites that took `*kernel-exception-lock*` to use the with-exception-lock macro.

In principle, there's nothing stopping a user from using *kernel-exception-lock* incorrectly, but someone poking around in the innards of CCL like this had better know what he is doing.

A new reproducer, suspend-spinlock-static-cons.lisp, shows the effects of the Lisp changes.

| platform | red (pre-change image) | green |
|---|---|---|
| darwinarm64 | wedged at cycle 0, 6.7M static conses | 500,000 cycles, 446,276,091 conses |
| linuxarm64 | wedged at 750 cycles, 9.7M conses | 500,000 cycles, 451,832,386 conses |

The linuxarm64 capture is the one that matters:

```
#0  sched_yield
#1  get_spin_lock (tcr=0xffff90000b70, p=0xaaaac686f0d0)          thread_manager.c:247
#2  unlock_recursive_lock_deferring_suspension (m=0xaaaac686f0a0) thread_manager.c:478
#3  unlock_exception_lock_in_handler (tcr=...)                    arm64-exceptions.c:2123
#4  signal_handler (signum=4, ...)                                arm64-exceptions.c:2248
--- threads by state ---  1 R, 25 S
```

Frame #2 is the C-side fix, executing, and wedged anyway because the word was
held by a frozen Lisp thread. That demonstrates that the C conversion alone was
not sufficient.

It also settles the question you reopened when you withdrew the test. You wrote
that you could no longer claim the by-design case — a suspended thread holding
the lock — was excluded. Frame #1 is `get_spin_lock`: it is the guard word, not
lock ownership. `p - m` is 0x30, the `spinlock` offset in `RECURSIVE_LOCK`, and
frame #3 says which lock.

I fixed %promote-rwlock in 0564225cd0b9. %promote-rwlock returning from its owner arm still holding the word is a real bug and is fixed, along with two (error :not-lock-owner ...) / (error :not-locked ...) calls in the same cond that pass keywords where condition names belong. Note your version released only under #-futex, leaving the futex build still holding.  (Not that we ever use #+futex at all.)

%unlock-rwlock-ptr posting semaphores under the spin word I am leaving alone. You are right that the futex twin releases first and that the non-futex one is the outlier, but the consequence is hold time rather than correctness: an rwlock holder that gets suspended is not deadlocked, so there is no window to close, only a longer one to shorten.

Although I didn't take your changes as written, your diagnosis was right.  I'll close the PR.  Thank you for the report and the reproducers.

